PR #7, opened 2016-10-25
bhgovconfig compiles

dabooz:


bmpotter:
Looks good, but 2 questions:

- Why did you comment out createConfigContent(). Is that done by some other part of the system now that the agbot is part of anax?
- Why did you delete the call to getHttpFile()? Doesn't that still need to be done?

dabooz:
1) Yes, because agbot is merged into anax now, agbot config will always be present for devmode cases. I commented it out because I am not sure if we will need to generate agbot configs for x86 agbot scenarios.
2) I didn't comment it out, it was already like that, I just removed the commented out code.

PR #21, opened 2016-11-02
added MTN_DEVICE_ID to container launch

michaeldye:


linggao:
Never mind about my comment above. Now I know how it works.

michaeldye:
I've change the set envvar to DEVICE_ID, an exception from the others, and we've created tickets to track fixing this in workloads.

PR #44, opened 2016-11-18
fixed the status page and details page to work with the new anax APIs.

linggao:


api/static/js/details.js
```diff
@@ -36,7 +36,7 @@ function fetchInfo(con_name) {
     }
 
     $.ajax({
-        url: '/contract',
+        url: '/agreement/all',
```
dabooz: How does this work? I see this code in the GET API:
		id := pathVars["id"]

		// we don't support getting just one yet
		if id != "" {
			w.WriteHeader(http.StatusBadRequest)
			return
		}
Does this pick up the /all and return an error?

PR #61, opened 2016-12-15
Feature/v2 ui

michaeldye:
You can find the source for the UI in https://github.com/michaeldye/anax-ui. The relationship between the projects is not yet solidified; for now, the ui is built there and the production artifacts are copied into this project.

Note that the woff files (fonts) are committed here b/c we prefer not to pull them from Google's CDN. These may end up in our object store someplace.


dabooz:
1) Why are there so many license files?
everything else looks ok. There's not much to review.

dabooz:
Answer is that these are licenses specific to font files imported from elsewhere that each have their own licenses, so the license is included with each font.

PR #84, opened 2017-01-12
Added v2 horizon api doc

linggao:


doc/api.md
```diff
@@ -1,230 +1,692 @@
...
+
+**Parameters:**
+
+Please refer to the response body of GET /horizondevice api.
+
 **Response:**
 
+code: 
+
+* 200 -- success
```
dabooz: The HTTP PATCH verb usually has a response code of 204? What do we do?

doc/api.md
```diff
@@ -1,230 +1,692 @@
...
+| services.{name}.attributes | array | an array of the user defined service attributes for the service. The attributes will be passed into the workloads environmental variables after the agreement is reached. |
+
+policy
+
+| name | type | description |
+| ---- | ---- | ---------------- |
+| header | json|  the header of the policy. It includes the name and the version of the policy. |
+| apiSpec | array | an array of api specifications. Each one includes a URL pointing to the definition of the API spec, the version of the API spec in OSGI version format, whether or not exclusive access to this API spec is required and the hardware architecture of the API spec implementation. | 
+| agreementProtocols | array | an array of agreement protocols. Each one includes the name of the agreement protocol.|
+| valueExchange | json | It includes the type of value exchange, the value being exchanged, the number of seconds between payments and a token used to identify the user of the value. |
```
dabooz: I would prefer to not document several of these fields until they actually mean something. Please omit valueExchange, resourceLimits, dataVerification, and proposalRejection. It might be good to also put in a statement like this "The policy also contains other fields that are unused and therefore not documented."

doc/api.md
```diff
@@ -1,230 +1,692 @@
...
+
+| name | type | description |
+| ---- | ---- | ---------------- |
+| attribute | json | Please refer to the response body for the GET /service/attribute api for the fields of an attribute.  |
+
+**Response:**
+
+code: 
+
+* 200 -- success
```
dabooz: This is probably 201?

doc/api.md
```diff
@@ -1,230 +1,692 @@
...
+none
+
+**Response:**
+
+code: 
+* 200 -- success
+
+body:
+
+| name | type | description |
```
dabooz: There are some new fields just added with my latest PR. Please add protocol_version to this doc. It's the version of the agreement protocol being used.

doc/api.md
```diff
@@ -1,230 +1,692 @@
...
+| ---- | ---- | ---------------- |
+| id   | string | the id of the agreement to be deleted. |
+
+**Response:**
+
+code: 
+
+* 200 -- success
```
dabooz: Usually DELETE returns 204, can you check this?

doc/api.md
linggao: Currently is still return 200 in anax api.

doc/api.md
linggao: Will do.

doc/api.md
linggao: yes, it is 201.

doc/api.md
linggao: Currently it is 200 in code.

doc/api.md
linggao: will do

PR #87, opened 2017-01-17
diagrams for agreement protocol

dabooz:


bmpotter:
Line 27 of horizonSequenceDiagram.txt should be "(4)", not "(2)", right?

Issue #64, opened 2016-12-19
archived agreements not sorted by the time


linggao:
The archived agreements returned by /agreeement API are not sorted by the timestamp which makes it very hard to understand what's going on if there are a lot of failure.

linggao:
This also causes the dashbord not showing the latest canceled the agreement for each service.

michaeldye:
This not resolved in the API, but there is a temporary UI-only fix in https://github.com/open-horizon/anax-ui/pull/3 .

linggao:
tested.

PR #92, opened 2017-01-27
Feature/sorted agreement api output

michaeldye:
Fixes issue #64


linggao:
very smart using the sort interface.

PR #130, opened 2017-02-27
The agreements that are being terminated but not yet archived are tre…

linggao:
…ated as archived in the /agreement api


dabooz:
Wow, that turned out to be really easy.

PR #142, opened 2017-03-09
Feature/eth container

dabooz:
Support ARM or X86 eth container....duh.
Also added support for deployment config signature verification, for eth containers and workload containers.


linggao:
I reviewed and approve this PR.

PR #143, opened 2017-03-09
agbot uses EthBlockchainEventMessage to dectect agreement creation an…

linggao:
…d termination on the blockchain instead of using rpc pulling.

### DIFF
--- a/agreementbot/agreementbot.go
+++ b/agreementbot/agreementbot.go
@@ -96,6 +96,14 @@ func (w *AgreementBotWorker) NewEvent(incoming events.Message) {
 			w.bcWritesEnabled = true
 		}
 
+	case *events.EthBlockchainEventMessage:
+		msg, _ := incoming.(*events.EthBlockchainEventMessage)
+		switch msg.Event().Id {
+		case events.BC_EVENT:
+			agCmd := NewBlockchainEventCommand(*msg)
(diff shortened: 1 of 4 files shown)


agreementbot/agreementbot.go
```diff
@@ -387,6 +400,44 @@ func (w *AgreementBotWorker) InitiateAgreementProtocolHandler(protocol string) {
...
 						glog.V(5).Infof("AgreementBot queued agreement cancellation")
 
+					case *BlockchainEventCommand:
+						glog.V(5).Infof("AgreementBot received blockchain event.")
+						cmd, _ := command.(*BlockchainEventCommand)
+
+						// Unmarshal the raw event
+						if rawEvent, err := protocolHandler.DemarshalEvent(cmd.Msg.RawEvent()); err != nil {
+							glog.Errorf("AgreementBotWorker unable to demarshal raw event %v, error: %v", cmd.Msg.RawEvent(), err)
+						} else if !protocolHandler.AgreementCreated(rawEvent) && !protocolHandler.ProducerTermination(rawEvent) && !protocolHandler.ConsumerTermination(rawEvent) {
```
dabooz: I think you can omit the check for !ConsumerTermination() because you're going to ignore that event anyway.

PR #151, opened 2017-03-17
added sequence diagrams for HA workload handling

michaeldye:


bmpotter:
- in workload-update diagram:
    - in Workload Update section, what does the arrow "negotiate agreement with Agbot" mean? The arrow above that established the agreement. I think instead you need the steps for verifying the data then making agreement with DeviceB
    - in the Error Case section:
        - if deviceA fails to successfully get back to version 1.1, what do you do with deviceB? Just keep running it in non-redundant mode?
        - if deviceA successfully gets back to version 1.1, do you then try to upgrade deviceB to 1.2?
- could you not put the png files in .gitignore? It would make it much more convenient for others to view the diagrams

michaeldye:
I've updated the PR with an update to the workload diagram source to clarify the update on Device A. Thanks for catching that the agreement negotiation for running workload version 1.2 on Device A is confusing, I've attempted to clarify that. Device B keeps running 1.1 at this point; this is the stable HA config while Device A gets the 1.2 update.

@dabooz and @datajerk should weigh in on your questions about behavior in the error case. I think a failure to roll back A to 1.1 should yield continually trying to launch 1.1 on A and leaving 1.1 on B until A is healthy. The answer to the second question is less clear to me: I think you leave B at 1.1 until you can successfully upgrade A to 1.2: the system has determined that 1.2 can't run on A, I think it should assume the workload is in error and not try deploying it.

I want to pull the entire public HA design / workload guarantees doc into this repo's wiki; when I do that I'll put the rendered PNGs in the doc. The google doc is useful now for collaborating on changes.

PR #198, opened 2017-05-01
Support device wide ha attribute

linggao:


api/api.go
```diff
@@ -845,6 +852,33 @@ func (a *API) serviceAttribute(w http.ResponseWriter, r *http.Request) {
...
+			// if the device is not HA enabled then the HA partner attribute is illegal
+			if !existingDevice.HADevice && attr.GetMeta().Id == "ha" {
+				glog.Errorf("Non-HA device %v does not support HA attribute", existingDevice)
+				writeInputErr(w, http.StatusBadRequest, &APIUserInputError{Input: "service.[attribute].Id", Error: "HA partner not permitted on non-HA devices"})
+				return
+			}
+
+			// Make sure that a device doesnt specify itself in the HA partner list
+			if existingDevice.HADevice {
+				if _, ok := attr.GetGenericMappings()["partnerID"]; ok {
```
dabooz: I think we should check that the attribute is an HA attribute before we check the Generic mapping. There could be another attribute int he future that uses the same partnerId keywords, but it would be a different attribute type. I made that same mistake in the service function. Can you correct it there too? I think the fix is just to check for attr.GetMeta().Id == "ha" before looking at the GenericMapping.

PR #201, opened 2017-05-04
Feature/agbot api

dabooz:
1) REST API for agbot agreements (including delete that will terminate agreements like the device side API)
2) archive terminated agreements in the agbot instead of deleting them so that info on old records can be retrieved
3) provide an auto delete feature that deletes archived agbot agreements (by default after 1 hour) so they dont build up in the database.

### DIFF
--- a/agreementbot/api.go
+++ b/agreementbot/api.go
@@ -52,7 +52,7 @@ func (a *API) listen(apiListen string) {
 	glog.Info("Starting AgreementBot API server")
 
 	// If there is no Agbot config, we will terminate
-	if a.Config.AgreementBot.APIListen == "" {
+	if apiListen == "" {
 		glog.Errorf("AgreementBotWorker API terminating, no AgreementBot API config.")
 		return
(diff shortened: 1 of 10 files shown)


agreementbot/api.go
```diff
@@ -0,0 +1,242 @@
...
+	}
+
+	return
+}
+
+func (a *API) listen(apiListen string) {
+	glog.Info("Starting AgreementBot API server")
+
+	// If there is no Agbot config, we will terminate
+	if a.Config.AgreementBot.APIListen == "" {
```
linggao: Is a.Config.AgreementBot.APIListen and the parameter apiListen the same thing?

dabooz: oh yeah, I didnt even notice that it was passed into the function. I will fix it in my next PR.

dabooz:
Added another commit to this PR that addresses your comment and adjusts the container worker running in the agbot in my e2edev environment.

PR #210, opened 2017-05-09
issue164 - dynamic policy in the agbot

dabooz:
Fixed issue 202 so that I could test.
Added unarchived filters to database searches in agbot. This was missed in the agbot/api feature.
Added a small bit of code to support multiple agreement protocols in the agbot. This was needed in order to properly react to a new policy with a new agreement protocol in it.
Most of the code is in support of dynamically adding/modifying policies int he agbot.

### DIFF
--- a/agreementbot/agreementbot.go
+++ b/agreementbot/agreementbot.go
@@ -32,7 +32,7 @@ type AgreementBotWorker struct {
 	protocols       map[string]bool
 	bc              *ethblockchain.BaseContracts
 	pm              *policy.PolicyManager
-	pwcommands      chan worker.Command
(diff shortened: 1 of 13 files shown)


dabooz:
Tests I executed before submitting the PR:
Unit tests in anax, of course.
In HA and non-HA scenarios, the following;
- change property value, workload priority, invalid workload signature, policy name, in agbot policy file
- rename policy file
- add policy file containing unsupported agreement protocol name
- add and delete policy file with unused device API
- delete and re-add policy file for device api that is being used
- touch policy file - i.e. no file content changes
- modify policy file so it is no longer valid json
- workload rollback

PR #222, opened 2017-05-16
disable workload rollback retry after successful data receipt'

dabooz:


agreementbot/wl_persistence.go
```diff
@@ -25,6 +25,8 @@ type WorkloadUsage struct {
     CurrentAgreementId string   `json:"current_agreement_id"`  // the agreement id currently in use
     FirstTryTime       uint64   `json:"first_try_time"`        // time when first agrement attempt was made, used to count retries per time
     LatestRetryTime    uint64   `json:"latest_retry_time"`     // time when the newest retry has occurred
+    DisableRetry       bool     `json:"disable_retries"`       // when true, retry and retry durations are disbled which effectively disables workload rollback
```
linggao: DisableRetry is singular but the json name disable_retries is not.

dabooz: ok, I will fix this today and send you and updated PR

PR #279, opened 2017-07-10
enable specific host and port for workloads

dabooz:
Enable workloads to specify a specific port to be bound on the host, and the host IP to bind to as well. Previously this capability was only available to infrastructure containers. We need this to get around a problem in a POC.

### DIFF
--- a/container/container.go
+++ b/container/container.go
@@ -76,7 +76,7 @@ type DeploymentDescription struct {
 }
 
 var invalidDeploymentOptions = map[string][]string {
-	"workload": []string{"Binds","SpecificPorts"},
+	"workload": []string{"Binds"},
 	"infrastructure": []string{},
 }
 


michaeldye:
Rather than expose specific ports for containers (which can easily cause port conflicts b/n patterns), the idea with the port mapping scheme in the container worker was to let docker pick the host port to use and map it to the predictable, in-container port as docker normally does. In this scheme, you can't predict the host's port, but you can discover it. There are two ways the port may need to be discovered:

1. To discover the host port from the host itself, you can use docker commands and / or port scanning commands (if you have a way to identify a service after connecting to it)

2. To discover the host port from an in-container pattern, we can write the chosen host port into the container's `workload_ro` directory, the location for sharing container pattern config with the container in the pattern

Will this strategy work for you?

dabooz:
Unfortunately it doesnt meet the POC need, that's why we had to do this.

michaeldye:
I'm guessing the PoC expects stable ports from the outside or something. The right way to do that is to use iptables prerouting rules to proxy the expected outside port to the discovered, random one. Can we do that?

bmpotter:
Mike,
1) this is a short term hack to get QP working quickly, won't necessarily last
2) why is iptables prerouting better?

michaeldye:
I get the quick need, Dave and I can get this working out of a PoC branch and later evaluate pulling the work in. That means we needn't go the distance on a feature like this, just the minimum.

On 2, the iptables way is just one way to map a stable port from the outside of the host to a random port on the host itself. The important part of the solution is that a workload doesn't expect to have a stable port number on the host b/c that could lead to conflicts. The iptables part would be a matter of static ops infrastructure config, Anax would do what it does now and set up dynamic network stuff that it tries hard to not have cause conflicts.

dabooz:
Closing this PR since we're going to do this in a temporary branch.

bmpotter:
But if 2 workloads want to listen on the same port on the host, the conflict would exist whether it is implemented the way dave is doing it, or with iptable mapping.

michaeldye:
That's right, an intended couple of patterns could have conflicting expectations for external ports. But in such a case, the ops-layer proxying (and possibly other networking config) could resolve it close to where the plumbing for that networking lives. It's the kind of abstraction docker makes in config: there is a "host config" set that describes how a container is configured w/r/t host-specific stuff and a separate "config" set that describes the container setup only and is portable.

bmpotter:
Ok, i see, thx.

PR #305, opened 2017-08-29
issue 303 - support workload config variables

dabooz:


governance/governance.go
```diff
@@ -957,6 +981,37 @@ func (w *GovernanceWorker) GetWorkloadPreference(url string) (map[string]string,
...
 
+// Get the environmental variables for the workload for MS split workloads. The workload config
+// record we have might not match the version of the workload being started so we will use the most
+// current config we have that is not newer than the workload version being started.
+func (w *GovernanceWorker) GetWorkloadConfig(url string, version string) (map[string]string, error) {
+
+	// Filter to return workload configs with versions less than or equal to the input workload version
+	OlderWorkloadWCFilter := func(workload_url string, version string) persistence.WCFilter {
+		return func(e persistence.WorkloadConfig) bool {
+			if e.WorkloadURL == workload_url && strings.Compare(e.Version, version) <= 0 {
```
linggao: What about version 2.0.1 vs. 2.0.01, should they be equal?

dabooz: Good question. As I said in slack, I looked at the OSGI spec and it is not clear on the whether or not leading zeroes are supported. Since our version system is closed, that is, doesnt need to interoperate outside our system, we can take the position that leading zeroes are illegal, and that will let us to lexigraphical comparisons on 2 version strings to determine equality, less than and greater than.  To do this, we can change the IsVersionString() method in version.go, and that will kick out any illegal strings before they get into the system.
Do you want to make the change to version.go or do you want me to do it. Doesnt matter to me.

linggao: Sure, I agree.  I will accept this PR and make the change in version.go.

PR #315, opened 2017-09-08
issue 295 - Automatic upgrade of microservices

linggao:


linggao:
Please do not merge yet.

api/api.go
```diff
@@ -556,9 +557,10 @@ func (a *API) service(w http.ResponseWriter, r *http.Request) {
 		var msdef *persistence.MicroserviceDefinition
 		if service.SensorVersion != nil {
 
-			// validate the version string
-			if !policy.IsVersionString(*service.SensorVersion) {
-				writeInputErr(w, http.StatusBadRequest, &APIUserInputError{Input: "sensor_version", Error: fmt.Sprintf("sensor_version %v is not a valid version string", *service.SensorVersion)})
+			// valideate the version string
```
dabooz: I dont think this is merged correctly. If if read it right, following this change we call Version_Expression_Factory again, I think the calls need to be consolidated.

PR #341, opened 2017-09-22
consolidate attributes of all types; reduce API function length; add tests

michaeldye:
*NOTE*: This merge cannot succeed until it is rebased on the current master's changes. It also creates breaking changes for registration code (although minor). I will handle this soon and @dabooz and I will merge it Monday AM and update the end-to-end tests.


linggao:
A new PR is in. Could you resolve the conflict files?

michaeldye:
@linggao: The changes in this PR have been rebased on the latest master's. Note that there are breaking changes here. They include the need to update reg_all payloads to use `type` instead of `short_type` and type names that match the struct names in https://github.com/michaeldye/anax/blob/refactor/consolidate_attributes/persistence/attribute_types.go (so `LocationAttributes` instead of `location`). Here's an example of an attribute body:

`{
  "type": "MappedAttributes",
  "label": "Application Mappings",
  "SensorUrls": ["https://foo", "https://goo"],
  "publishable":	false,
  "mappings": {
    "KEY": "VALUE"
  }
}`

(See https://github.com/michaeldye/anax/blob/refactor/consolidate_attributes/api/api_int_test.go for examples). POST, PUT, PATCH, DELETE are supported now with ID fields.

Note that the path for standalone attribute registration is not any longer `/service/attribute` but `/attribute`. You'll have to adjust stuff like this: https://github.com/linggao/horizon-utils/blob/master/reg/reg_all#L397.

@dabooz Please have a look at the api.go changes to make sure I haven't missed any of the org work you did today. I carefully checked the merge but it was pretty big so I could use more eyes. Thanks!

PR #351, opened 2017-09-28
fix policy lib test problems

dabooz:


policy/workload.go
```diff
@@ -31,19 +31,9 @@ func (self *WorkloadList) Add_Workload(new_ele *Workload) error {
...
-	Signature string `json:"signature"`
-}
-
-func (i Image) IsSame(compare Image) bool {
-	return i.File == compare.File && i.Signature == compare.Signature
-}
-
 type Torrent struct {
 	Url       string `json:"url,omitempty"`
-	Signature string `json:"signature:"signature,omitempty"`
```
michaeldye: ah, good catch

PR #364, opened 2017-10-09
Issue 355 - autoconfigure microservices that dont have variables to set

dabooz:
Requires Exchange 1.38.
Requires changes to reg_all and any doc due to the need to use the new configstate API in order to make a node ready to make agreements.
Design doc: https://docs.google.com/document/d/1ZclxhHZP7BcBESGcx5oAGUxyew2aj1vZ0ffRXeX-vXQ
Please note that ONLY phase 0 of that design is implemented in this PR.

### DIFF
--- a/agreement/agreement.go
+++ b/agreement/agreement.go
@@ -323,6 +323,11 @@ func (w *AgreementWorker) handleDeviceRegistered(cmd *DeviceRegisteredCommand) {
 		}
 	}
 
+	// Do a quick sync to clean out old agreements in the exchange.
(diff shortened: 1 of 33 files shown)


dabooz:
The second commit in the PR has the following:
- additional unit tests, we could do more but this gets us to 20% coverage of the API
- node side anax will delete its own agreements in the exchange if you wipe out its database and then restart it. It does not (cannot) delete the corresponding agbot side of the agreement.
- restructure some additional APIs, not all are done yet and added some unit tests for those.

PR #365, opened 2017-10-10
issue 313 - Anax update exchange with container status

linggao:
Requires exchange 1.37.0

### DIFF
--- a/governance/governance.go
+++ b/governance/governance.go
@@ -48,6 +48,7 @@ type GovernanceWorker struct {
 	devicePattern string
 	pm            *policy.PolicyManager
 	producerPH    map[string]producer.ProducerProtocolHandler
+	deviceStatus  *DeviceStatus
 }
 
 func NewGovernanceWorker(cfg *config.HorizonConfig, db *bolt.DB, pm *policy.PolicyManager) *GovernanceWorker {
@@ -80,6 +81,7 @@ func NewGovernanceWorker(cfg *config.HorizonConfig, db *bolt.DB, pm *policy.Poli
 		deviceToken:   token,
 		devicePattern: pattern,
 		producerPH:    make(map[string]producer.ProducerProtocolHandler),
+		deviceStatus:  NewDeviceStatus(),
 	}
 
 	worker.start()
@@ -115,6 +117,9 @@ func (w *GovernanceWorker) NewEvent(incoming events.Message) {
 			cmd := w.NewCleanupStatusCommand(msg.AgreementProtocol, msg.AgreementId, STATUS_WORKLOAD_DESTROYED)
 			w.Commands <- cmd
 		}
+
+		cmd := w.NewReportDeviceStatusCommand()
+		w.Commands <- cmd
 
 	case *events.TorrentMessage:
 		msg, _ := incoming.(*events.TorrentMessage)
@@ -176,6 +181,7 @@ func (w *GovernanceWorker) NewEvent(incoming events.Message) {
 			cmd := producer.NewExchangeMessageCommand(*msg)
 			w.Commands <- cmd
 		}
+
 	case *events.ContainerMessage:
 		msg, _ := incoming.(*events.ContainerMessage)
 		if msg.LaunchContext.Blockchain.Name == "" { // microservice case
@@ -187,7 +193,14 @@ func (w *GovernanceWorker) NewEvent(incoming events.Message) {
 				cmd := w.NewUpdateMicroserviceCommand(msg.LaunchContext.Name, false, 1, "Failed to launch containers.")
 				w.Commands <- cmd
 			}
+
+			cmd := w.NewReportDeviceStatusCommand()
+			w.Commands <- cmd
 		}
+	case *events.MicroserviceContainersDestroyedMessage:
+
+		cmd := w.NewReportDeviceStatusCommand()
+		w.Commands <- cmd
 
 	default: //nothing
 	}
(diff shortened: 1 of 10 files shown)


governance/status.go
```diff
@@ -0,0 +1,413 @@
...
+
+// Report the containers status and connectivity status to the exchange.
+func (w *GovernanceWorker) reportStatus() {
+
+	go func() {
+
+		// This is the amount of time for the governance routine to wait.
+		waitTime := w.Config.Edge.ReportDeviceStatusInterval
+		if waitTime == 0 { // never report status when the interval is set to 0
+			glog.V(5).Info("ReportDeviceStatusInterval is 0. The status report to the exchange is turned off.")
```
dabooz: This should be an Infof without the verbose setting, just glog.Infof(... because we always want this message to come out in the log.

governance/status.go
```diff
@@ -0,0 +1,413 @@
...
+			if wl_status, err := w.getWorkloadStatus(); err != nil {
+				glog.Errorf(logString(fmt.Sprintf("Error getting microservice container status: %v", err)))
+			} else {
+				device_status.Workloads = wl_status
+			}
+
+			//compare the device status with the one gotten from last time
+			if jbytes, err := json.Marshal(&device_status); err != nil {
+				glog.V(5).Infof(logString(fmt.Sprintf("Failed to convert the device status %v to json: %v", device_status, err)))
+			} else if Device_Status != nil && device_status.IsSame(*Device_Status) {
```
dabooz: You wrote a lot of IsSame and comparison code, which is all very similar and repetitive. Wouldnt it have been easier/better to marshal the status to a byte array and then hash that, and then compare the hash from last time to see if anything has changed? Less duplicate code to maintain, and might even be faster. The reflection APIs (DeepEqual) are known to be slow.

governance/status.go
```diff
@@ -0,0 +1,413 @@
...
+				glog.V(5).Infof(logString(fmt.Sprintf("device status is the same as last time. No need to report: %v", string(jbytes[:]))))
+			} else {
+
+				Device_Status = &device_status
+				glog.V(5).Infof(logString(fmt.Sprintf("device status to report to the exchange: %v", string(jbytes[:]))))
```
dabooz: Dont log the byte array of what we're sending to the exchange, it is very hard to debug from that, log the objects and let the String() functions format them in the log.

governance/status.go
```diff
@@ -0,0 +1,413 @@
...
+	// Filter to return all instances for a msdef
+	msdefFilter := func(msdef_id string) persistence.MIFilter {
+		return func(a persistence.MicroserviceInstance) bool {
+			return a.MicroserviceDefId == msdef_id
+		}
+	}
+
+	status := make([]MicroserviceStatus, 0)
+
+	if client, err := docker.NewClient(w.Config.Edge.DockerEndpoint); err != nil {
```
dabooz: If you factor this docker call out of this API, then you can (a) only call it once, and (b) I think getMicroserviceStatus and getWorkloadStatus and getContainerStatus will become unittestable functions. It is hard to mock the docker call so you can't UT these large functions.

governance/status.go
linggao: Done

governance/status.go
linggao: The status contains slices and the element order in them may be different. So we cannot use hash to compare them.

governance/status.go
linggao: It is not bytes. It is actually shown as json structure in the log file. I purposely added this here so that I know what it will look like in the exchange.

governance/status.go
linggao: I do not understand the comment. Need to discuss this with you.

governance/governance.go
```diff
@@ -176,6 +181,10 @@ func (w *GovernanceWorker) NewEvent(incoming events.Message) {
 			cmd := producer.NewExchangeMessageCommand(*msg)
 			w.Commands <- cmd
 		}
+
```
dabooz: This is the exchange message event, why do you generate the report just because there is a message from an agbot?

governance/status.go
```diff
@@ -0,0 +1,305 @@
...
+	}
+
+	// get workload status
+	if wl_status, err := w.getWorkloadStatus(containers); err != nil {
+		glog.Errorf(logString(fmt.Sprintf("Error getting microservice container status: %v", err)))
+	} else {
+		device_status.Workloads = wl_status
+	}
+
+	//compare the device status with the one gotten from last time
```
dabooz: we arent comparing any more

governance/status.go
```diff
@@ -0,0 +1,305 @@
...
+		// report the status to the exchange
+		if err := w.writeStatusToExchange(&device_status); err != nil {
+			glog.Errorf(logString(err))
+		}
+	}
+
+}
+
+// Find the status for all the microservices
+func (w *GovernanceWorker) getMicroserviceStatus(containers []docker.APIContainers) ([]MicroserviceStatus, error) {
```
dabooz: We need unit tests for getMicroserviceStatus and getWorkloadStatus just like you have for getContainerStatus.

governance/governance.go
linggao: Good catch. It meant to put where the device id and token were assigned.  I have deleted it.

governance/status.go
linggao: fixed the comment.

governance/status.go
dabooz: We decided to defer those to an issue, Ling is opening it.

PR #387, opened 2017-10-25
Issue 385 - Restore default for MS privileged setting to false

linggao:


dabooz:
You need to verify with bruce that any msdefs which need this privileged flag flag have it set in the deployment string, same for any blockchain definitions in the exchange.

PR #415, opened 2017-11-08
anax API doc update

linggao:


doc/api.md
```diff
@@ -171,6 +171,69 @@ curl -s -w "%{http_code}" -X PATCH -H 'Content-Type: application/json'  -d '{
...
+
+* 204 -- success
+
+body: 
+
+none
+
+**Example:**
+```
+curl -s -w "%{http_code}" -X DELETE -H 'Content-Type: application/json'  -d '{
```
dabooz: No, block and removeNode are not passed in the body, they are passed as query parameters.
curl -X DELETE http://localhost/horizondevice?block=true&removeNode=false

PR #461, opened 2017-11-30
Feature/clean remove cli

michaeldye:
Carl and I like this.

PR #463, opened 2017-12-01
small hzn changes and enhancements

bmpotter:
- remove token display from show node, instead of displaying null
- remove exclusiveAccess from show services output (because it is not getting set properly)
- change show sub-cmds to cmds like 'hzn keys list'
- delete 1 agreement or all agreements
- create user/pw if doesn't exist and in public org


michaeldye:
Please merge Makefile changes with https://github.com/open-horizon/anax/pull/461; the others look good.

bmpotter:
Ok, i used your Makefile changes in PR 461 instead of mine, because yours were better.

Merging now so we can get this in staging.

PR #522, opened 2018-01-19
feature/cli list certs

michaeldye:
API docs were updated, `/trust?verbose=true` was made a little more predictable, and the CLI can list trusted certs and their attributes (using `hzn key list`).

See attached doc for some detailed use examples including exporting a cert and public key from the API and using the rsapss tool to verify a signature.

[hzn_and_api_key_and_cert_examples.txt](https://github.com/open-horizon/anax/files/1644898/hzn_and_api_key_and_cert_examples.txt)

### DIFF
--- a/api/api_publickey.go
+++ b/api/api_publickey.go
@@ -35,7 +35,7 @@ func (a *API) publickey(w http.ResponseWriter, r *http.Request) {
 				errorHandler(NewSystemError(fmt.Sprintf("Error returning content of %v/%v, error %v", resource, fileName, err)))
 			}
 		} else {
-			if out, err := FindPublicKeysForOutput(a.Config, verbose != ""); err != nil {
+			if out, err := FindPublicKeysForOutput(a.Config, verbose == "true"); err != nil {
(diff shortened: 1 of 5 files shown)


bmpotter:
Mike,

- I see in your examples the output for `hzn key list` for certs. What does the output look like for non-wrapped raw keys?
- So this is a change to the anax api, because `/publickey` is changed to `/trust`, right? I don't think we have anyone relying on `/publickey` (except for hzn, which you changed), but i just wanted to highlight it here so everyone is aware.

michaeldye:
1. There is no way to see a bare RSA PSS key in output from `hzn key list`, just the `trust/` API directly without the verbose flag. I intended using bare RSA PSS keys a feature only support to prevent breakage of WIoTP and so working with those keys through the `hzn key list` isn't something I wanted to provide. Do you think we need more support for it?

2. The API supports both `publickey/` and `trust/` paths. The only published path is `trust/`, hopefully we can eliminate `publickey/` soon.

PR #563, opened 2018-02-27
 Issue 498 - Add signing keys to ms/wk/pat exchange definition 

linggao:


governance/microservice.go
```diff
@@ -306,6 +331,7 @@ func (w *GovernanceWorker) RollbackMicroservice(msdef *persistence.MicroserviceD
 			return nil
 		} else if err := w.UpgradeMicroservice(msdef, new_msdef, false); err != nil {
 			glog.Errorf(logString(fmt.Sprintf("Failed to downgrade %v from version %v key %v to version %v key %v. %v", msdef.SpecRef, msdef.Version, msdef.Id, new_msdef.Version, new_msdef.Id, err)))
+			msdef = new_msdef
```
dabooz: Is this an unrelated change?

producer/producer_protocol_handler.go
```diff
@@ -145,8 +147,12 @@ func (w *BaseProducerProtocolHandler) HandleProposal(ph abstractprotocol.Protoco
 		handled = true
 	} else if tcPolicy, err := policy.DemarshalPolicy(proposal.TsAndCs()); err != nil {
 		glog.Errorf(BPPHlogString(w.Name(), fmt.Sprintf("received error demarshalling TsAndCs, %v", err)))
+	} else if err := w.saveSigningKeys(tcPolicy); err != nil {
```
dabooz: I'm not sure how to fix this, so I'm not sure if you should fix it, but I would like to make an observation. You are saving the signing keys early in this function so that you can do the Is_Self_Consistent check with all the keys. That makes sense. However, the side effect of saving early is that if the proposal gets rejected later in this function, then you will have a public key on the node that is essentially extra garbage, which will never get cleaned up. Could this become a security exposure? I am not sure how it could, but I have to ask anyway.

governance/microservice.go
linggao: Yes, it is unrelated, but I found the problem while testing the signing keys.

producer/producer_protocol_handler.go
linggao: You are right, cleaning up the keys is a problem.  I do not think it is a security exposure.

PR #564, opened 2018-02-27
Issue 557 - agbot policy manager failed initializing when there are 2…

linggao:
… policies with the same name

### DIFF
--- a/policy/policy_file.go
+++ b/policy/policy_file.go
@@ -657,6 +657,22 @@ func (c *Contents) GetPolicyName(org, filename string) string {
 	return ""
 }
 
+// check if there is already a tracked policy file with a policy name that is the same as the new policy that we're trying to add.
+// It returns the name of the conflict file.
+func (c *Contents) ConflictsWithAlreadyTracked(org string, pol *Policy) string {
+	if !c.HasOrg(org) {
+		return ""
+	} else {
+		for fn, we := range c.AllWatches[org] {
+			if we.Pol.Header.Name == pol.Header.Name {
+				return fn
+			}
+		}
+
+	}
+	return ""
+}
+
 func CreatePolicyFile(filepath string, org string, name string, p *Policy) (string, error) {
 
 	// Store the policy on the filesystem in an org based hierarchy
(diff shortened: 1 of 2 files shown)


policy/policy_manager_test.go
```diff
@@ -52,13 +52,6 @@ func Test_Payloadmanager_init_success2(t *testing.T) {
 	}
 }
 
-func Test_Payloadmanager_dup_policy_name(t *testing.T) {
-
-	if _, err := Initialize("./test/pfduptest/", make(map[string]string), nil, true); err == nil {
```
dabooz: It's ok that you removed this test but you should replace it with a test of the new function.

policy/policy_manager.go
```diff
@@ -175,8 +175,11 @@ func Initialize(policyPath string, arch_synonymns config.ArchSynonyms, workloadR
 	// Setup the callback functions for the policy file watcher. The change notification is the only callback
 	// that should be invoked at this time.
 	changeNotify := func(org string, fileName string, policy *Policy) {
-		numberFiles += 1
-		pm.AddPolicy(org, policy)
+		if err := pm.AddPolicy(org, policy); err != nil {
+			glog.Errorf("Unable to add policy file %v/%v during initialization, error: %v", org, fileName, err)
+		} else {
```
dabooz: You dont need these changes any more with the new code in PolicyFileChangeWatcher

policy/policy_file.go
```diff
@@ -657,6 +657,24 @@ func (c *Contents) GetPolicyName(org, filename string) string {
 	return ""
 }
 
+// check if another file in the same org has the same policy name as another file.
+// It returns the name of the conflict file.
+func (c *Contents) ConflictsWith(org, filename string, pol *Policy) string {
+	if !c.HasOrg(org) {
+		return ""
+	} else if c.HasFile(org, filename) {
```
dabooz: You dont need this check. The caller has already checked that policy file is not being tracked.

policy/policy_manager_test.go
linggao: Will do.

policy/policy_manager.go
linggao: It is do guard other error, just in case.

policy/policy_file.go
linggao: There are 2 callers for PolicyFileChangeWatcher(), one is from the policy_manager and the other is from agreementbot. PolicyFileChangeWatcher uses this check to make sure the fileChanged function is not get called and this file is not put into the watcher map.

PR #600, opened 2018-04-11
Issue 593 - agbot failed to initialize when the org for the pattern i…

linggao:
…s deleted

### DIFF
--- a/agreementbot/agreementbot.go
+++ b/agreementbot/agreementbot.go
@@ -196,6 +196,13 @@ func (w *AgreementBotWorker) Initialize() bool {
 		return false
 	}
 
+	// To strart clean, remove all left over pattern based policy files.
+	// This is only called once at the agbot start up time
+	if err := policy.DeleteAllPolicyFiles(w.BaseWorker.Manager.Config.AgreementBot.PolicyPath, true); err != nil {
+		glog.Errorf("AgreementBotWorker cannot clean up pattern based policy files under $v. %v", w.BaseWorker.Manager.Config.AgreementBot.PolicyPath, err)
+		return false
+	}
+
 	// Give the policy manager a chance to read in all the policies. The agbot worker will not proceed past this point
 	// until it has some policies to work with.
 	for {
@@ -1105,6 +1112,7 @@ func (w *AgreementBotWorker) GeneratePolicyFromPatterns() int {
 		return -1
 	}
 
+	glog.V(5).Infof(AWlogString(fmt.Sprintf("pattern manager initialized: %v", w.PatternManager)))
 	return 0
 }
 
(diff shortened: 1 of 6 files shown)


policy/policy_file.go
```diff
@@ -867,18 +940,25 @@ func DeleteAllPolicyFiles(homePath string) error {
...
+			}
 		}
 	}
 	return nil
 }
 
 // This is an internal function used to find all policy files in the policy directory. Files that don't end in
 // .policy are ignored. Directories are also ignored.
-func getPolicyFiles(homePath string) ([]os.FileInfo, error) {
+func GetPolicyFiles(homePath string) ([]os.FileInfo, error) {
```
dabooz: I would rather not make this function available outside the policy package. It looks like it doesnt need to be capitalized.

agreementbot/agreementbot.go
```diff
@@ -1129,12 +1137,25 @@ func (w *AgreementBotWorker) internalGeneratePolicyFromPatterns() error {
...
-			return errors.New(fmt.Sprintf("unable to get patterns for org %v, error %v", org, err))
+		var exchangePatternMetadata map[string]exchange.Pattern
+		var err error
+
+		// check if the org exists on the exchange or not
+		if _, err = exchange.GetOrganization(w.Config.Collaborators.HTTPClientFactory, org, w.GetExchangeURL(), w.GetExchangeId(), w.GetExchangeToken()); err != nil {
+			// org does not exist is returned as an error
+			glog.V(5).Infof(AWlogString(fmt.Sprintf("unable to get organization %v: %v", org, err)))
+			exchangePatternMetadata = make(map[string]exchange.Pattern)
+			glog.V(5).Infof(AWlogString(fmt.Sprintf("unable to get organization len %v", len(exchangePatternMetadata))))
```
dabooz: Do we need this log entry?

agreementbot/pattern_manager.go
```diff
@@ -192,7 +192,32 @@ func (pm *PatternManager) UpdatePatternPolicies(org string, definedPatterns map[
 
 	// Exit early on error
 	if !pm.hasOrg(org) {
-		return errors.New(fmt.Sprintf("org %v not found in pattern manager", org))
+		return nil
```
dabooz: why is this not an error any more?

policy/policy_file.go
linggao: yes, my mistake. Forgot to switch it back.

agreementbot/pattern_manager.go
linggao: I added the error back because the UpdatePatternPolicies is only called when there the org is in the pm.

agreementbot/agreementbot.go
linggao: deleted the last debug message.

PR #643, opened 2018-06-13
nil pointer dereference fix

tlammi:
In container/container.go function NewContainerLaunchContext()
is called with nil as last parameter which causes nil pointer
dereference.

This commit changes the stored value to pointer to fix this issue.

### DIFF
--- a/events/events.go
+++ b/events/events.go
@@ -176,7 +176,7 @@ type ContainerLaunchContext struct {
 	Name                 string // used as the docker network name and part of container name. For microservice it is the ms instance key
 	AgreementId          string
 	Microservices        []MicroserviceSpec                     // Service dependencies go here. Microservices (in the workload/microservice model) never have dependencies.
-	ServicePathElement   *persistence.ServiceInstancePathElement // The service that we're trying to start.
+	ServicePathElement   persistence.ServiceInstancePathElement // The service that we're trying to start.
 }
 
 func (c ContainerLaunchContext) String() string {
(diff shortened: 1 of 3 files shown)


events/events.go
```diff
@@ -176,7 +176,7 @@ type ContainerLaunchContext struct {
 	Name                 string // used as the docker network name and part of container name. For microservice it is the ms instance key
 	AgreementId          string
 	Microservices        []MicroserviceSpec                     // Service dependencies go here. Microservices (in the workload/microservice model) never have dependencies.
-	ServicePathElement   persistence.ServiceInstancePathElement // The service that we're trying to start.
+	ServicePathElement   *persistence.ServiceInstancePathElement // The service that we're trying to start.
```
linggao: Please do not change it to pointer because pointer in the event structure may cause segmentation fault where passing from a thread to another.

events/events.go
```diff
@@ -200,24 +200,19 @@ func (c ContainerLaunchContext) GetMicroservices() []MicroserviceSpec {
...
-	c.Blockchain = bc
-	c.Name = name
-	c.AgreementId = agId
-	c.Microservices = mss
-	if(spe != nil){
-		c.ServicePathElement = new(persistence.ServiceInstancePathElement)
-		*c.ServicePathElement = *spe
-	} else {
-		c.ServicePathElement = nil
+	return &ContainerLaunchContext{
```
linggao: Could you add code here to guard against spe to be nil? 
When spe is nil, make `spe_temp := persistence.NewServiceInstancePathElement("", "")` and `c.ServicePathElement  = *spe_temp`.

PR #691, opened 2018-10-05
Issue 29A - Anax support using pattern in different org

linggao:


agreementbot/agreementbot.go
```diff
@@ -797,14 +797,22 @@ func DeleteMessage(msgId int, agbotId, agbotToken, exchangeURL string, httpClien
...
 // microservices.
-func (w *AgreementBotWorker) searchExchange(pol *policy.Policy, searchOrg string) (*[]exchange.SearchResultDevice, error) {
+func (w *AgreementBotWorker) searchExchange(pol *policy.Policy, polOrg string) (*[]exchange.SearchResultDevice, error) {
 
 	// If it is a pattern based policy, search by worload URL and pattern.
 	if pol.PatternId != "" {
+		// get a list of node orgs that agbot is serving for this pattern
+		nodeOrgs := w.PatternManager.GetServedNodeOrgs(polOrg, exchange.GetId(pol.PatternId))
+		if len(nodeOrgs) == 0 {
+			glog.V(3).Infof("Policy file for pattern %v exits but currently the agbot is not serving this policy for any organizations.", pol.PatternId)
```
dabooz: typo: "pattern %v exits", should be "exists"

agreementbot/agreementbot.go
```diff
@@ -797,14 +797,22 @@ func DeleteMessage(msgId int, agbotId, agbotToken, exchangeURL string, httpClien
...
 // by pattern. If the agbot is working with a manually created policy file, then it will do searches by list of
 // microservices.
-func (w *AgreementBotWorker) searchExchange(pol *policy.Policy, searchOrg string) (*[]exchange.SearchResultDevice, error) {
+func (w *AgreementBotWorker) searchExchange(pol *policy.Policy, polOrg string) (*[]exchange.SearchResultDevice, error) {
 
 	// If it is a pattern based policy, search by worload URL and pattern.
 	if pol.PatternId != "" {
+		// get a list of node orgs that agbot is serving for this pattern
+		nodeOrgs := w.PatternManager.GetServedNodeOrgs(polOrg, exchange.GetId(pol.PatternId))
+		if len(nodeOrgs) == 0 {
```
dabooz: If the agbot is not configured to serve any specific node orgs, then isnt it always serving the same org that the pattern is in, by default?

agreementbot/pattern_manager.go
```diff
@@ -128,8 +135,57 @@ func (pm *PatternManager) hasPattern(org string, pattern string) bool {
...
+	pm.spMapLock.Lock()
+	defer pm.spMapLock.Unlock()
+
+	// copy the input map
+	pm.ServedPatterns = servedPatterns
+}
+
+// chek if the agbot serves the given pattern or not.
+func (pm *PatternManager) servePattern(pattern_org string, pattern string) bool {
+	for _, sp := range pm.ServedPatterns {
```
dabooz: I think you need the spMapLock here to protect against write while you are reading. There may be other places too.

api/path_node.go
```diff
@@ -56,6 +56,7 @@ func FindHorizonDeviceForOutput(db *bolt.DB) (*HorizonDevice, error) {
 	var device *HorizonDevice
 
 	pDevice, err := persistence.FindExchangeDevice(db)
+	glog.Infof("*** the device is: %v", pDevice)
```
dabooz: Do we really need this?

agreementbot/agreementbot.go
bmpotter: I'm unclear about how we get to this situation. I did tell ling that when running the `/orgs/{orgid}/agbots/{id}/patterns` exchange api that if they omit `nodeOrgids` in the body, the exchange will automatically set it to the pattern org. But if they explicitly set `"nodeOrgids": [ ]` i currently don't change it, i just save that. Is this the case we have here?

linggao: This is not the case where nodeOrgid is omitted or empty in the `POST /orgs/{orgid}/agbots/{id}/patterns` call. It is the case when the user has a pattern but forgot to call `POST/orgs/{orgid}/agbots/{id}/patterns` for this pattern.

agreementbot/pattern_manager.go
linggao: servePattern() is in the same same process with the setServedPatterns(). The only thing that runs in a different process is GetServedNodeOrgs() which has spMapLock already.  Thoug it is safe to add spMapLock in servePattern() and servePattern() just in case someday someone will cal them from a different process. So I will add spMapLock for them.

agreementbot/agreementbot.go
linggao: fixed.

api/path_node.go
linggao: oops, debug code. will remove.

agreementbot/agreementbot.go
bmpotter: I'm confused, if the user didn't call `POST/orgs/{orgid}/agbots/{id}/patterns`, how does the agbot even know about this pattern and have a policy file for it?

linggao: It could be a left over from last time when deletion failed. It could be user copied it manually. Our code has to guard for this cases.

bmpotter: In an offline discussion ling clarified to me that the `if len(nodeOrgs) == 0 {` should never happen, because `PatternManager.GetServedNodeOrgs()` will default `nodeOrgs` to the pattern org. But `if len(nodeOrgs) == 0 {` is here as a safeguard anyway so that this code doesn't have to know about that subtle behavior inside `PatternManager.GetServedNodeOrgs()`. That's a good practice, so i'm resolving this conversation.

PR #698, opened 2018-10-10
Issue 683 - remove workloads and microservices

dabooz:


cli/hzn.go
```diff
@@ -133,67 +132,13 @@ Environment Variables:
 	exPatDelCmd := exPatternCmd.Command("remove", "Remove a pattern resource from the Horizon Exchange.")
 	exDelPat := exPatDelCmd.Arg("pattern", "The pattern to remove.").Required().String()
 	exPatDelForce := exPatDelCmd.Flag("force", "Skip the 'are you sure?' prompt.").Short('f').Bool()
-	exPatternAddWorkCmd := exPatternCmd.Command("insertworkload", "Add or replace a workload in an existing pattern resource in the Horizon Exchange.")
```
linggao: Do we want to support `hzn pattern insert/remove service`?

dabooz: Bruce and I discussed it. He told me that the insert workload command was a hack for wiotp, and that we wouldn't need it for services.

PR #706, opened 2018-10-24
Issue 687 - Use docker volumes for service persistence

linggao:
Please review the comment of issue 687 for design details.

### DIFF
--- a/container/container.go
+++ b/container/container.go
@@ -379,14 +379,14 @@ func CreateCLIContainerWorker(config *config.HorizonConfig) (*ContainerWorker, e
 
 func NewContainerWorker(name string, config *config.HorizonConfig, db *bolt.DB) *ContainerWorker {
 
-	storage_base_dir := "/var/tmp/horizon/service_storage" // default
+	// if config.Edge.ServiceStorage is not empty, then we assume that the local file system directory will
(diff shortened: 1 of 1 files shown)


container/container.go
```diff
@@ -946,23 +948,27 @@ func (b *ContainerWorker) ResourcesCreate(agreementId string, agreementProtocol
...
-	// write raw to workloadRWStorageDir
-	if err := ioutil.WriteFile(path.Join(workloadRWStorageDir, "Configure"), configureRaw, 0644); err != nil {
-		return nil, err
+		glog.V(5).Infof("Writing raw config to file in %v. Config data: %v", workloadRWStorageDir, string(configureRaw))
+		// write raw to workloadRWStorageDir
+		if err := ioutil.WriteFile(path.Join(workloadRWStorageDir, "Configure"), configureRaw, 0644); err != nil {
+			return nil, err
+		}
+	} else {
+		// do we need to create the volume here?
```
dabooz: Where do you create the volume?

linggao: I need to remove the `else` clause here. For the volume case, we do nothing here. The volume is created by docker as long as the source in `source:dest:option` in the HostConfig.Binds is a name instead of a file system path and the named volume does not exist.

PR #770, opened 2019-01-29
Issue 769 - Correct the documentation of the specific_ports deploymen…

linggao:
…t field

### DIFF
--- a/cli/native_deployment/native_deployment.go
+++ b/cli/native_deployment/native_deployment.go
@@ -135,7 +135,7 @@ func (p *NativeDeploymentConfigPlugin) Validate(dep interface{}) (bool, error) {
 }
 
 // This can't be a const because a map literal isn't a const in go
-var VALID_DEPLOYMENT_FIELDS = map[string]int8{"image": 1, "privileged": 1, "cap_add": 1, "environment": 1, "devices": 1, "binds": 1, "specific_ports": 1, "command": 1, "ports": 1}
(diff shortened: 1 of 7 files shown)


cli/native_deployment/native_deployment.go
```diff
@@ -135,7 +135,7 @@ func (p *NativeDeploymentConfigPlugin) Validate(dep interface{}) (bool, error) {
 }
 
 // This can't be a const because a map literal isn't a const in go
-var VALID_DEPLOYMENT_FIELDS = map[string]int8{"image": 1, "privileged": 1, "cap_add": 1, "environment": 1, "devices": 1, "binds": 1, "specific_ports": 1, "command": 1, "ports": 1}
+var VALID_DEPLOYMENT_FIELDS = map[string]int8{"image": 1, "privileged": 1, "cap_add": 1, "environment": 1, "devices": 1, "binds": 1, "ports": 1, "command": 1, "ephemeral_ports": 1}
```
dabooz: If we still support specific_ports then it has to be in this list, doesnt it?

linggao: We encourage the new user use the `ports`. The `specific_ports` is for backward compatibility now. I think people who use `hzn dev` are developing new code, hence they should switch to use `ports`.

TheMosquito: The `hzn dev` command is one of our interfaces, so like David Booz, I think this is an important place to ensure backward compatibility.  I have been helping David Martin to use `hzn dev` commands in his CI/CD process, for example.

PR #819, opened 2019-03-08
Issue 817 - Handle new style service url name properly in agbot in te…

linggao:
…rms of policy generating
1) Fixed the way the policy file name is generated for agbot.
2) Changed a few service urls to not using the "http://" format in e2edev test.
3) Changed org name e2edev to e2edev@somecomp.com in e2edev test.


exchange/rpc.go
```diff
@@ -833,15 +833,19 @@ func GetPatterns(httpClientFactory *config.HTTPClientFactory, org string, patter
 // Create a name for the generated policy that should be unique within the org.
 func makePolicyName(patternName string, workloadURL string, workloadOrg string, workloadArch string) string {
 
-	url := ""
+	url := workloadURL
+
+	// for the old url style, get the part after '*://'
 	pieces := strings.SplitN(workloadURL, "/", 3)
+	fmt.Printf("pieces=%v\n", pieces)
```
dabooz: This needs to be a glog call if you want to keep it.

PR #827, opened 2019-03-12
Issue 786 - 'hzn exchange' command should support -n flag

linggao:


cli/exchange/node.go
```diff
@@ -14,13 +14,42 @@ type ExchangeNodes struct {
...
+func getCred(userPw string, nodeIdTok string) string {
+	credToUse := ""
+
+	if userPw != "" {
+		credToUse = userPw
+	} else {
+		if nodeIdTok != "" {
+			credToUse = nodeIdTok
+		} else {
+			if tmpU := cliutils.WithDefaultEnvVar(&userPw, "HZN_EXCHANGE_USER_AUTH"); *tmpU != "" {
```
bmpotter: HZN_EXCHANGE_USER_AUTH is already read in hzn.go and passed in as userPw if set, so there is no need to check it here.

cli/hzn.go
```diff
@@ -312,10 +315,15 @@ Environment Variables:
...
+	cliutils.Verbose("Full command: %s", fullCmd)
 	if strings.HasPrefix(fullCmd, "exchange") {
 		exOrg = cliutils.RequiredWithDefaultEnvVar(exOrg, "HZN_ORG_ID", "organization ID must be specified with either the -o flag or HZN_ORG_ID")
-		exUserPw = cliutils.RequiredWithDefaultEnvVar(exUserPw, "HZN_EXCHANGE_USER_AUTH", "exchange user authentication must be specified with either the -u flag or HZN_EXCHANGE_USER_AUTH")
+
+		// hzn exchange node list command can take either -u user:pw or -n nodeid:token as credentials. It will be handled in the NodeList function
+		// other hzn exchange commands take -u user:pw as credentials
+		if !strings.Contains(fullCmd, "node list") && !strings.Contains(fullCmd, "node settoken") && !strings.Contains(fullCmd, "node remove") && !strings.Contains(fullCmd, "node confirm") {
+			exUserPw = cliutils.RequiredWithDefaultEnvVar(exUserPw, "HZN_EXCHANGE_USER_AUTH", "exchange user authentication must be specified with either the -u flag or HZN_EXCHANGE_USER_AUTH")
+		}
```
bmpotter: If you are doing all the checking for the cmds that can have node auth, why not also read HZN_EXCHANGE_NODE_AUTH here? That keeps the logic together and will make getCred() in node.go simpler. Also, since i'd also like you to also support node auth for service list and pattern list, probably best to put all of the cmds that can accept node auth in an array, so it will be more obvious to that you need to add to that array if a new sub-cmd is added that can take node auth. Thx!

cli/exchange/node.go
linggao: In this case it is not read from hzn.go. The one that read in hzn.go are for other commands that will exit if no userPW or HZN_EXCHANGE_USER_AUTH are specified. Here we allow it to be empty.

cli/hzn.go
linggao: Yes, if we are going to support node auth for service list, pattern list etc. It makes sense to move the getCred here.

PR #834, opened 2019-03-21
Issue 833 - node messaging key became empty string after rebooting

linggao:


agreement/agreement.go
```diff
@@ -171,6 +171,10 @@ func (w *AgreementWorker) Initialize() bool {
 	}
 
 	if w.GetExchangeToken() != "" {
+		// populate the privFileName and pubFileName variables in the exchange.messaging.go
+		if _, _, err := exchange.GetKeys(""); err != nil {
+			glog.Error("Failed to get the messaging keys. %v", err)
```
dabooz: Please use the logString wrapper for error msgs and try to follow the same syntax for the msgs. Look at the error handling for the syncOnInit call for an example.

PR #864, opened 2019-04-09
policy dynamicity

dabooz:


governance/governance.go
```diff
@@ -374,39 +388,93 @@ func (w *GovernanceWorker) governAgreements() {
...
-						w.Messages() <- events.NewGovernanceWorkloadCancelationMessage(events.AGREEMENT_ENDED, events.AG_TERMINATED, ag.AgreementProtocol, ag.CurrentAgreementId, ag.GetDeploymentConfig())
-						// clean up microservice instances if needed
-						w.handleMicroserviceInstForAgEnded(ag.CurrentAgreementId, false)
+						w.cancelGovernedAgreement(&ag,
+							fmt.Sprintf("Start terminating agreement for %v. Reason: %v", ag.RunningWorkload.URL, w.producerPH[ag.AgreementProtocol].GetTerminationReason(reason)),
+							reason,
+							persistence.EC_CANCEL_AGREEMENT_EXECUTION_TIMEOUT)
+					}
+				} else {
+					// Finalized agreements could become out of policy if the policy changes on the node. Verify that the existing agreement
```
linggao: Can the following part be put in a function?

dabooz: yes, I agree. I think it needs to be part of the agreement protocol code, but since I know we will be revisting this we can clean it up then.

PR #901, opened 2019-04-24
 Issue 846 - Add config file support to hzn

linggao:


cli/dev/hznenv.go
```diff
@@ -1,33 +1,19 @@
...
 
 import (
-	"fmt"
+	"github.com/open-horizon/anax/cli/cliconfig"
 )
 
-const HZNENV_FILE = "hzn.cfg"
+const HZNENV_FILE = "hzn.json"
 
 // It creates a hzn.env file that contains the enviromental variables needed for this project.
```
bmpotter: Comment is misleading because the code doesn't create hzn.env

cli/dev/image.go
```diff
@@ -16,7 +16,7 @@ const DEFAULT_BASE_OS_VALUE = "alpine:latest"
 const MAKE_FILE_CONTENT = `# Make targets for building the IBM example helloworld edge service
 
 # This imports the variables from horizon/hzn.cfg. You can ignore these lines, but do not remove them.
--include horizon/.hzn.cfg.tmp.mk
+-include PROJECT_DIR/.hzn.cfg.tmp.mk
```
bmpotter: Also a nit, but the file should now be called `.hzn.json.tmp.mk`

cli/hzn.go
```diff
@@ -45,9 +46,13 @@ Environment Variables:
       to specify the organization ID'.
   HZN_EXCHANGE_USER_AUTH:  Default value for the 'hzn exchange -u' or 'hzn
       register -u' flag, in the form '[org/]user:pw'.
-  HZN_DONT_SUBST_ENV_VARS:  Set this to "1" to indicate that input json files
```
bmpotter: Is the idea behind removing this from the help that we will still support it as a hidden option, but almost no one will need to know about it, so don't bother confusing them with it?

linggao: You are absolutely right.

cli/dev/image.go
linggao: done

cli/dev/hznenv.go
linggao: done

Issue #926, opened 2019-05-10
Adding 'hzn exchange' commands for business policy


linggao:
hzn exchange business listpolicy [<policy_id>] 
hzn exchange business addpolicy  <policy_id>  -f <json_file_name>
hzn exchange business removepolicy <policy_id> 
hzn exchange business updatepolicy  <policy_id>  ... (patch a policy)

LiilyZhang:
Verified the about 4 commands with args on icp-staging, hzn version `2.23.10`

PR #931, opened 2019-05-16
Issue #926 added 'hzn exchange' commands for business policy

MaxMcAdam:


cli/exchange/business.go
```diff
@@ -0,0 +1,146 @@
...
+}
+
+//BusinessAddPolicy will add a new policy or overwrite an existing policy byt he same name in the Horizon Exchange
+func BusinessAddPolicy(org string, credToUse string, policy string, jsonFilePath string) {
+	cliutils.SetWhetherUsingApiKey(credToUse)
+	org, credToUse = cliutils.TrimOrg(org, credToUse)
+
+	//read in the new business policy from file
+	newBytes := cliconfig.ReadJsonFileWithLocalConfig(jsonFilePath)
+	var policyFile exchange.ExchangeBusinessPolicy
```
linggao: Please use businesspolicy.BusinessPolicy structure here instead of exchange.ExchangeBusinessPolicy.

cli/exchange/business.go
```diff
@@ -0,0 +1,146 @@
...
+		if err != nil {
+			cliutils.Fatal(cliutils.JSON_PARSING_ERROR, "failed to unmarshal json input file %s: %v", valueFilePath, err)
+		}
+		err = newValue.Validate()
+		if err != nil {
+			cliutils.Fatal(cliutils.CLI_INPUT_ERROR, "Invalid format for properties")
+		}
+		patch[attribute] = newValue
+		cliutils.ExchangePutPost(http.MethodPatch, cliutils.GetExchangeUrl(), "orgs/"+org+"/business/policies"+cliutils.AddSlash(policy), cliutils.OrgAndCreds(org, credToUse), []int{201}, patch)
+		fmt.Println("Policy " + org + "/" + policy + " updated in the Horizon Exchange")
```
linggao: The patch should take more than one attributes at a time. You do not have to break them up. This is my understanding. Can you confirm it with bp?

linggao: Currently it only accept the first attribute in the input. Will need to get back to it once supporting multiple changes.

PR #944, opened 2019-05-22
Issue 892 - added validation for external policy type properties

MaxMcAdam:


externalpolicy/property_list.go
```diff
@@ -86,13 +151,89 @@ func (self *PropertyList) Add_Property(new_ele *Property) error {
...
+			}
+		case float64, json.Number:
+			if declaredType == INTEGER_TYPE && fmt.Sprintf("%T", actualType) == "float64" {
+				if float64(int(property.Value.(float64))) != property.Value.(float64) {
+					return fmt.Errorf("Value %v of property %s is not an intger type", property.Value, property.Name)
+				}
+			} else if declaredType == INTEGER_TYPE && fmt.Sprintf("%T", actualType) == "json.Number" {
+				_, err := property.Value.(json.Number).Int64()
+				if err != nil {
+					return fmt.Errorf("Value %v of property %s is not an intger type", property.Value, property.Name)
```
dabooz: integer type, his typo exists in a few places

externalpolicy/property_list_test.go
```diff
@@ -51,6 +52,46 @@ func Test_PropertyList_compatible(t *testing.T) {
...
+	p2 = `[{"name":"prop2","value":"val2"},{"name":"prop1","value":"1.2.53","type":"version"}]`
+	if pl1 = create_PropertyList(p1, t); pl1 != nil {
+		if pl2 = create_PropertyList(p2, t); pl2 != nil {
+			if err := pl1.Compatible_With(pl2); err != nil {
+				t.Errorf("Error: %v is compatible with %v, error was %v\n", p1, p2, err)
+			}
+		}
+	}
+
+	p1 = `[{"name":"prop1","value":["a","b","c"],"type":"list of string"},{"name":"prop2","value":"val2"}]`
```
dabooz: list of string is actually supposed to be comma separated values in 1 string, not a proper json array

externalpolicy/property_list.go
```diff
@@ -86,13 +151,89 @@ func (self *PropertyList) Add_Property(new_ele *Property) error {
...
+	for _, property := range *self {
+		if property.Name == "" || property.Value == nil {
+			return fmt.Errorf("Property must include a name and a value: %v", property)
+		}
+		declaredType := property.Type
+		if declaredType == "" {
+			declaredType = STRING_TYPE
+		}
+		switch actualType := property.Value.(type) {
+		case bool:
```
dabooz: I would like the validation to be smarter. It should be ok to specify a boolean property value without the type field. When properties are compared the (runtime) type of the value has to be checked to ensure they are the same.

Issue #892, opened 2019-04-23
Implement text based constraint language processor


dabooz:
Depends on the plugin created by (https://github.com/open-horizon/anax/issues/893)
- [x] Support version, integer, float, string, boolean and list of string types (Lily)
- [x] Support boolean operators and aliases (&&, ||, etc) in the constraint language (Lily)
- [x] Output JSON according to the ./policy/counter_party_properties.go schema (Max)
- [x] Node policy always contains the builtin node side properties, including patching the policy in the exchange if the policy does not contain the builtins. (Ling)
- [x] Implement validation of properties in external policy type - may need to add an optional Type field to the PropertyList type in order to distinguish version from string, etc. (Max)
- [x] GetNextExpression and GetNextOperator return correct result for values with spaces. (Max))
- [x] Property support  version, integer, float, string, boolean and list of string types. Give error in Add_Property, Validate and IsSame functions.(Max)

PR #948, opened 2019-05-23
Issue #892 point 3. Format required properties from constraints

MaxMcAdam:


policy/counter_party_properties.go
```diff
@@ -54,79 +54,81 @@ func RequiredPropertyFromConstraint(extConstraint *externalpolicy.ConstraintExpr
 	var nextExpression, remainder, controlOp string
 	var handler plugin_registry.ConstraintLanguagePlugin
 
-	// Completely empty input gets a completely empty, but non-nil response.
-	newRP := RequiredProperty_Factory()
-	if extConstraint == nil || len(*extConstraint) == 0 {
```
dabooz: We need to keep the nil guard for extConstraint

dabooz: accidentally approved

externalpolicy/counter_party_properties_test.go
```diff
@@ -686,6 +686,86 @@ func Test_merge4(t *testing.T) {
...
+	(*ce) = append((*ce), "prop == value && prop2 == value2", "prop3 == value3 || prop4 == value4 && prop5 == value5", "prop6 >= val6 || prop7 < val7")
+	if rp, err := RequiredPropertyFromConstraint(ce); err != nil {
+		t.Errorf("Error: unable to convert simple expression: %v", err)
+	} else if tle := rp.TopLevelElements(); tle == nil {
+		t.Errorf("Error: There should be 3 top level array elements")
+	} else if len(tle) != 3 {
+		t.Errorf("Error: Should be 1 top level array element")
+	}
+}
+
```
linggao: Can you use the following as a test case? It is from test_language_test.go 
````
    constraintStrings := []string{
        "iame2edev == true && cpu == 3 || memory <= 32",
        "hello == \"world\"",
        "hello in \"'hi world', 'test'\"",
        "eggs == \"truck load\" AND certification in \"USDA, Organic\"",
        "version == 1.1.1 OR USDA == true",
        "version in [1.1.1,INFINITY) OR cert == USDA",
    }
````
You need to add support for new type version and string array and also string 'in' operator in counterparty_propteties.go. And please use IsSatisfiedBy() function from constraint_expression.go to validate the conversion.

PR #964, opened 2019-05-30
Issue #892  - validate input properties and support all prope…

MaxMcAdam:
…rty types


MaxMcAdam:
- Added support for integer types in Validation

- Added test cases with the types supported by constraints

- Added validation in Add_Property

- Added support for list of string type properties that I had forgotten before

PR #968, opened 2019-06-03
Issue 945 - MMS object placement

dabooz:


agreementbot/object_policy.go
```diff
@@ -0,0 +1,248 @@
...
+
+		}
+	}
+
+	return
+
+}
+
+// Get node policy
+func (w *AgreementBotWorker) GetNodePolicy(deviceId string) (*policy.Policy, error) {
```
linggao: There is a same function in BaseAgreementWorker.  Should we create one without class name instead?

dabooz: Yes, that's exactly what I will do in my next PR. There is a bug in the sync service code that is preventing me from finishing that section of event handling code. When that bug is resolved, I will be able to know if I still need that function to be used by both the main worker and the protocol workers.

PR #993, opened 2019-06-06
Issue 898 - Remove Use of Policy Attributes (from Previous Policy Des…

linggao:
…ign)


test/gov/apireg.sh
```diff
@@ -173,47 +139,8 @@ fi
...
     if [ "$PATTERN" = "sns" ] || [ "$PATTERN" = "spws" ]; then
 
-        read -d '' pwsservice <<EOF
-{
-  "url": "https://bluehorizon.network/services/no-such-service",
-  "organization": "e2edev@somecomp.com",
-  "version": "1.0.0",
-  "attributes": [
-    {
-      "type": "ComputeAttributes",
```
dabooz: We need to keep this test so you need to change the attribute to something other than compute, like userinput.

PR #1016, opened 2019-06-14
Issue 793 - support versionRange in service def

dabooz:


exchange/service.go
```diff
@@ -173,6 +174,21 @@ func (w *GetServicesResponse) ShortString() string {
...
 
+// The version field of all service dependencies is being changed to versionRange for all external
+// interactions. Internally, since we still have to support old service defs, we will copy the
+// new field into the old field for backward compatibility. This function make an in place
+// modification of itself.
+func (w *GetServicesResponse) SupportVersionRange() {
+	for id, sdef := range w.Services {
+		for ix, sdep := range sdef.RequiredServices {
+			if sdep.Version == "" {
+				w.Services[id].RequiredServices[ix].Version = w.Services[id].RequiredServices[ix].VersionRange
```
linggao: Should it be the other way?

PR #1023, opened 2019-06-16
Fixed pattern case so that it can be independent of UserInputAttributes.

linggao:


test/gov/apireg.sh
```diff
@@ -190,73 +190,168 @@ fi
 
 # =======================================================================
 # Setup services/workloads
-
-echo -e "\nNo netspeed setting is $NONS"
-if [ "$NONS" != "1" ]
+if [ "$NO_USERINPUT_ATTRIBUTES" != "1" ]
```
dabooz: Instead of configuring all services one way or the other, I think you should config some with service config API and some with userinput API, then both paths get tested all the time.

linggao: Good idea.

PR #1045, opened 2019-06-19
Convert CSS container to run as non-root

ctiijima:
Runs the CSS in the container as a non-root user.

note: This will require the css to listen on a port above 1024 as this is a requirement to run as non-root.

### DIFF
--- a/css/image/cloud-sync-service-amd64/Dockerfile
+++ b/css/image/cloud-sync-service-amd64/Dockerfile
@@ -1,7 +1,12 @@
 FROM alpine:3.6
 
-RUN apk --no-cache add libcrypto1.0 libssl1.0 ca-certificates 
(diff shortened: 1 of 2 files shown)


css/image/cloud-sync-service-amd64/Dockerfile
```diff
@@ -1,7 +1,12 @@
 FROM alpine:3.6
 
-RUN apk --no-cache add libcrypto1.0 libssl1.0 ca-certificates 
+RUN addgroup -g 1000 -S cssuser && adduser -u 1000 -S cssuser -G cssuser \
+        && apk --no-cache add libcrypto1.0 libssl1.0 ca-certificates
```
dabooz: Need to update the anax/Makefile with the new version/tag of the CSS if you want the build system to create a new container.

ctiijima: The make file has been updated to css version 1.0.11.

Issue #1065, opened 2019-06-24
hzn dev service new must create patterns non-public by default


dabooz:
the exchange now only allows public patterns in the IBM org.

jiportilla:
tested this fix on ICP via VM, works as intended

-rw-rw-r-- 1 ivanp ivanp  487 Jun 25 15:22 pattern.json

PR #1067, opened 2019-06-24
Issue #1065 - hzn dev service new must create patterns non-public by …

MaxMcAdam:
…default


cli/dev/servicedefinition.go
```diff
@@ -41,7 +41,7 @@ func CreateServiceDefinition(directory string, specRef string, imageInfo map[str
 
 	res.Org = "$HZN_ORG_ID"
 	res.Label = "$SERVICE_NAME for $ARCH"
-	res.Public = true
+	res.Public = false
```
dabooz: This is in the wrong place. It's the pattern that has to be non-public

PR #1070, opened 2019-06-25
Issue 960 - disallow root binds

dabooz:


container/container.go
```diff
@@ -2260,3 +2283,30 @@ func isAnaxNetwork(net *docker.Network, bridgeName string) bool {
 	}
 	return false
 }
+
+// Verify that the permission bits for the host side of the binding allow anyone/other
+// to access that file or directory.
+func hasValidBindPermissions(binds []string) error {
+	for _, bind := range binds {
+		containerVol := strings.Split(bind, ":")
+		if len(containerVol) > 1 && containerVol[1] != "" {
```
linggao: We should cover the case where there is only 1 element. like "/mydir". This means both host and docker container has the same dir name.

container/container.go
```diff
@@ -2260,3 +2283,30 @@ func isAnaxNetwork(net *docker.Network, bridgeName string) bool {
...
+
+			info, err := os.Stat(containerVol[0])
+			if err != nil {
+				return err
```
linggao: There are a few cases where is should not return an error:
1. The item is a volume name which docker will create later.
2. The item is a directory name but the directory does not exist, docker will create it.
Here it should changed to `continue` instead of `return err`.

container/container.go
dabooz: ok i wondering about that case.

container/container.go
dabooz: Is there a way to discern those cases from a real error?

PR #1083, opened 2019-07-08
Issue 1074 -  hzn userinput add/update should verify the user input before saving it

LiilyZhang:
When service does not exist in the exchange:
Screen Shot 2019-07-08 at 16 15 15
Screen Shot 2019-07-08 at 16 15 29
Screen Shot 2019-07-08 at 16 16 13

When given an userinput variable that doesn't define in the service, show a warning message in log but still save the userinput:
Screen Shot 2019-07-08 at 16 16 29

When given a wrong type of userinput variables, show errors:
Screen Shot 2019-07-08 at 16 18 56
Screen Shot 2019-07-08 at 16 19 24

Support for list of strings:
Screen Shot 2019-07-09 at 11 54 45
Screen Shot 2019-07-09 at 11 54 54

Warning log updated:
Screen Shot 2019-07-09 at 11 55 45

### DIFF
--- a/api/path_node_userinput.go
+++ b/api/path_node_userinput.go
@@ -241,7 +241,7 @@ func isCorrectType(policyInputValue interface{}, serviceInputType string) (bool,
 		if isFloat64(policyInputValue) {
 			return true, nil
 		}
-	case "boolean":
+	case "boolean", "bool":
 		if isBoolean(policyInputValue) {
 			return true, nil
 		}
--- a/api/path_node_userinput.go
+++ b/api/path_node_userinput.go
@@ -4,11 +4,14 @@ import (
 	"errors"
 	"fmt"
 	"github.com/boltdb/bolt"
+	"github.com/golang/glog"
 	"github.com/open-horizon/anax/events"
 	"github.com/open-horizon/anax/exchange"
 	"github.com/open-horizon/anax/exchangesync"
 	"github.com/open-horizon/anax/persistence"
 	"github.com/open-horizon/anax/policy"
+	"github.com/open-horizon/anax/semanticversion"
+	"strings"
 )
 
 // Return an empty user input object or the object that's in the local database.
@@ -28,6 +31,7 @@ func UpdateNodeUserInput(userInput []policy.UserInput,
 	errorhandler DeviceErrorHandler,
 	getDevice exchange.DeviceHandler,
 	patchDevice exchange.PatchDeviceHandler,
+	getService exchange.ServiceHandler,
 	db *bolt.DB) (bool, []policy.UserInput, []*events.NodeUserInputMessage) {
 
 	// Check for the device in the local database. If there are errors, they will be written
@@ -37,6 +41,17 @@ func UpdateNodeUserInput(userInput []policy.UserInput,
 		return errorhandler(nil, NewSystemError(fmt.Sprintf("Unable to read node object, error %v", err))), nil, nil
 	} else if pDevice == nil {
 		return errorhandler(nil, NewNotFoundError("Exchange registration not recorded. Complete account and node registration with an exchange and then record node registration using this API's /node path.", "node")), nil, nil
+	}
+
+	// verify userinput: 1) service exist, 2) variables definied in the service, otherwise return true but give warning message 3) values have correct type
+	validated := false
+	for _, u := range userInput {
+		validated, err = ValidateUserInput(u, getService)
+		if !validated {
+			return errorhandler(nil, NewSystemError(fmt.Sprintf("Failed to validate node userInput, error: %v", err))), nil, nil
+		} else if err != nil { // validate == true, but give back warning error message
+			glog.Warningf(logString(fmt.Sprintf("Warning message: Userinput validated: %v. error: %v \n", validated, err)))
+		}
 	}
 
 	if changedSvcs, err := exchangesync.UpdateNodeUserInput(pDevice, db, userInput, getDevice, patchDevice); err != nil {
@@ -54,6 +69,7 @@ func PatchNodeUserInput(patchObject []policy.UserInput,
 	errorhandler DeviceErrorHandler,
 	getDevice exchange.DeviceHandler,
 	patchDevice exchange.PatchDeviceHandler,
+	getService exchange.ServiceHandler,
 	db *bolt.DB) (bool, []policy.UserInput, []*events.NodeUserInputMessage) {
 
 	pDevice, err := persistence.FindExchangeDevice(db)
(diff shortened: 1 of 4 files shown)


api/path_node_userinput.go
```diff
@@ -115,3 +142,148 @@ func DeleteNodeUserInput(errorhandler DeviceErrorHandler, db *bolt.DB,
 	nodeUserInputUpdated := events.NewNodeUserInputMessage(events.UPDATE_NODE_USERINPUT, *chnagedSvcSpecs)
 	return false, []*events.NodeUserInputMessage{nodeUserInputUpdated}
 }
+
+var logString = func(v interface{}) string {
+	return fmt.Sprintf("path_node_userinput %v", v)
```
dabooz: You can get rid of this function, and use the apiLogString function.

api/path_node_userinput.go
```diff
@@ -63,6 +79,17 @@ func PatchNodeUserInput(patchObject []policy.UserInput,
...
 	}
 
+	// verify userinput: 1) service exist, 2) variables definied in the service, otherwise return true but give warning message 3) values have correct type
+	validated := false
+	for _, u := range patchObject {
+		validated, err = ValidateUserInput(u, getService)
+		if !validated {
+			return errorhandler(nil, NewSystemError(fmt.Sprintf("Failed to validate node userInput, error: %v", err))), nil, nil
+		} else if err != nil { // validate == true, but give back warning error message
+			glog.Warningf(logString(fmt.Sprintf("Warning message: Userinput validated: %v. error: %v \n", validated, err)))
```
dabooz: Change to glog.Warningf(apiLogString(fmt.Sprintf("PATCH node/userinput %v ", err)))

api/path_node_userinput.go
```diff
@@ -39,6 +43,17 @@ func UpdateNodeUserInput(userInput []policy.UserInput,
...
 	}
 
+	// verify userinput: 1) service exist, 2) variables definied in the service, otherwise return true but give warning message 3) values have correct type
+	validated := false
+	for _, u := range userInput {
+		validated, err = ValidateUserInput(u, getService)
+		if !validated {
+			return errorhandler(nil, NewSystemError(fmt.Sprintf("Failed to validate node userInput, error: %v", err))), nil, nil
+		} else if err != nil { // validate == true, but give back warning error message
+			glog.Warningf(logString(fmt.Sprintf("Warning message: Userinput validated: %v. error: %v \n", validated, err)))
```
dabooz: Change to glog.Warningf(apiLogString(fmt.Sprintf("UPDATE node/userinput %v ", err)))

api/path_node_userinput.go
```diff
@@ -115,3 +142,148 @@ func DeleteNodeUserInput(errorhandler DeviceErrorHandler, db *bolt.DB,
...
+	glog.V(3).Infof(logString(fmt.Sprintf("Start validate userinput .... \n")))
+	serviceOrg := userInput.ServiceOrgid
+	serviceUrl := userInput.ServiceUrl
+	serviceVersionRange := userInput.ServiceVersionRange
+	serviceArch := userInput.ServiceArch
+
+	// The versionRange field is checked for valid characters by the Version_Expression_Factory, it has a very
+	// specific syntax and allows a subset of normally valid characters.
+
+	// Use a default sensor version that allows all version if not specified.
```
dabooz: default service version. And same change below

api/path_node_userinput.go
```diff
@@ -115,3 +142,148 @@ func DeleteNodeUserInput(errorhandler DeviceErrorHandler, db *bolt.DB,
...
+	if err != nil {
+		errorString = fmt.Sprintf("versionRange %v cannot be converted to a version expression, error %v", userInput.ServiceVersionRange, err)
+		return false, errors.New(errorString)
+	}
+
+	// service exist
+	var sdef *exchange.ServiceDefinition
+	sdef, _, err = getService(serviceUrl, serviceOrg, vExp.Get_expression(), serviceArch)
+	if sdef == nil {
+		errorString = fmt.Sprintf("Service not exist for org: %v, url: %v, version: %v, arch: %v, get error: %v \n", serviceOrg, serviceUrl, vExp.Get_expression, serviceArch, err)
```
dabooz: Service does not exist ....

api/path_node_userinput.go
```diff
@@ -115,3 +142,148 @@ func DeleteNodeUserInput(errorhandler DeviceErrorHandler, db *bolt.DB,
...
+
+	if len(inputNameNotDefinedInService) != 0 {
+		errorString = fmt.Sprintf("The following variables are not defined in userInput for service in its highest version: %v \n", inputNameNotDefinedInService)
+		return true, errors.New(errorString)
+	}
+
+	return true, nil
+}
+
+func isCorrectType(policyInputValue interface{}, serviceInputType string) (bool, error) {
```
dabooz: We support "list of strings" as a type also

test/gov/gpstest_apireg.sh
```diff
@@ -13,7 +13,7 @@ echo -e "Pattern is set to $PATTERN"
     {
       "serviceOrgid": "IBM",
       "serviceUrl": "https://bluehorizon.network/service-gps",
-      "serviceArch": "",
+      "serviceArch": "amd64",
```
dabooz: This can be left empty by the user, so we want it to stay that way in the tests. Same goes for the change to PWS.

api/path_node_userinput.go
```diff
@@ -115,3 +142,148 @@ func DeleteNodeUserInput(errorhandler DeviceErrorHandler, db *bolt.DB,
...
+	vExp, err := semanticversion.Version_Expression_Factory(serviceVersionRange)
+	var errorString string
+	if err != nil {
+		errorString = fmt.Sprintf("versionRange %v cannot be converted to a version expression, error %v", userInput.ServiceVersionRange, err)
+		return false, errors.New(errorString)
+	}
+
+	// service exist
+	var sdef *exchange.ServiceDefinition
+	sdef, _, err = getService(serviceUrl, serviceOrg, vExp.Get_expression(), serviceArch)
```
dabooz: If serviceArch is empty at this point, you should use cutil.ArchString() to get the node's arch so that you are retrieving the correct service for this node's arch. And now that I think about it, if serviceArch is not empty then it has to be the same as cutil.ArchString().

api/path_node_userinput.go
LiilyZhang: Added check for list of strings

api/path_node_userinput.go
LiilyZhang: Update:
1. Allows leave empty for serviceArch
2. Add a check for serviceArch = nodeArch if serviceArch is not empty, will return false if not match

Issue #1089, opened 2019-07-10
update --help for hzn CLI


dabooz:
There are several places to update:
1) The -u flag for hzn exchange --help, as an alternative to using -o, the user can set HZN_ORG_ID
2) The HZN_EXCHANGE_USER_AUTH env var description in hzn --help, include text to say that HZN_ORG_ID can be set if org is omitted when HZN_EXCHANGE_USER_AUTH is set.
3) Examine the rest of the CLI to see if there are any more places where -o (as a user org - I think there place where -o is used as a service org so those places are irrelevant to this issue) is mentioned without mentioning HZN_ORG_ID.

jiportilla:
[] 1. current hzn exchange --help messages for -o & -u flag are: 

-o, --org=ORG          The Horizon exchange organization ID. If not specified, HZN_ORG_ID will be used as a default.

-u, --user-pw=USER:PW  Horizon Exchange user credentials to query and create exchange resources. If not specified, HZN_EXCHANGE_USER_AUTH will be used as a default. If you don’t prepend it with the user’s org, it will automatically be prepended with the -o value.

The updated message for hzn exchange --help -u flag shall be:

-u, --user-pw=USER:PW  Horizon Exchange user credentials to query and create exchange resources. If not specified, HZN_EXCHANGE_USER_AUTH will be used as a default. If you don’t prepend it with the user’s org, it will automatically be prepended with the -o value. As an alternative to using -o, you can set HZN_ORG_ID with the Horizon exchange organization ID.

jiportilla:
() 2 Current help message for HZN_EXCHANGE_USER_AUTH in hzn --help is:

HZN_EXCHANGE_USER_AUTH:  Default value for the 'hzn exchange -u' or 'hzn
      register -u' flag, in the form '[org/]user:pw'.

The updated message for HZN_EXCHANGE_USER_AUTH shall be:

HZN_EXCHANGE_USER_AUTH:  Default value for the 'hzn exchange -u' or 'hzn
      register -u' flag, in the form '[org/]user:pw'. *Notice that HZN_ORG_ID can be set if org is omitted when HZN_EXCHANGE_USER_AUTH is set.

jiportilla:
/[ ] # '/[ ]' 3 . HZN_EXCHANGE_USER_AUTH is listed in help msg for hzn exchange -- usr-pw , hzn register, dev (dev service start, dev service validate, dev dependency fetch) and hzn mms

[...]

PR #1091, opened 2019-07-12
issue #1089 update hzn CLI --help

jiportilla:


cli/hzn.go
```diff
@@ -72,7 +72,7 @@ Environment Variables:
 
 	exchangeCmd := app.Command("exchange", "List and manage Horizon Exchange resources.")
```
dabooz: You also need to modify the text describing the HZN_EXCHANGE_USER_AUTH env var.

jiportilla: updated HZN_EXCHANGE_USER-AUTH environment variable, ran all tests successfully

Issue #1072, opened 2019-06-26
Add certificate argument to horizon-container or /etc/default/horizon


bmpotter:
It is awkward to use an ICP self-signed cert with the anax container. After starting the container, you have to copy the cert into the container, run `update-ca-certificates` in the container, and then restart anax in the container (not restart the container).

Since self-signed cert is the primary use case for ICP, we should make this more convenient. I can think of 2 options:

1. Add an arg to `horizon-container` to pass the self-signed cert
1. Since this is also a need for regular anax (on the host), altho it is not as inconvenient, we could add a variable that is supported in `/etc/default/horizon` that could point to the cert file, and have both anax in container and non-container anax honor that.

Option 2 is my preference.

PR #1092, opened 2019-07-15
Issue #1072 - Add certificate argument to /etc/default/horizon

MaxMcAdam:
Unrelated changes in `api/path_node_userinput.go` are from `make format`

### DIFF
--- a/anax-in-container/horizon-container
+++ b/anax-in-container/horizon-container
@@ -120,8 +120,12 @@ checkRequirements() {
 # Pull the latest horizon docker image, get some prereqs established, and start the image
 # First arg is optional host default file, second arg has value 'updating' if we are called by restart
 start() {
+	#variables sourced from the default file are available in this script but not its children
 	if [[ -n "$1" ]]; then
 		defaultFileMountArg="-v ${1}:/etc/default/horizon:ro"
+		source $1
+	else
+		source /etc/default/horizon
 	fi
 
 	checkRequirements	# this will exit with msg if requirements are not met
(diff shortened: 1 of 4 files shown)


anax-in-container/horizon-container
```diff
@@ -122,6 +122,9 @@ checkRequirements() {
 start() {
 	if [[ -n "$1" ]]; then
 		defaultFileMountArg="-v ${1}:/etc/default/horizon:ro"
+		source /etc/default/horizon
+	else
```
bmpotter: this else is when `$1` is not defined, so why does it source `$1` ?

bmpotter: also, you should add a comment to the `source` statements that says that it makes the variables in `/etc/default/horizon` available to *this* script, but none of its children (because the variables in the file aren't exported). That won't be obvious to future developers modifying this code.

bmpotter: oh, also, in the case in which `$1` is set, it should source `$1` not `/etc/default/horizon`. (The latter is the location in the container, and we are still on the host.)

bmpotter: so the net is that this section should be:```
	if [[ -n "$1" ]]; then
		defaultFileMountArg="-v ${1}:/etc/default/horizon:ro"
                 source "$1"
	fi
```

File: anax-in-container/horizon-container [hunk 2]
```diff
@@ -181,6 +184,13 @@ start() {
 		checkrc $? "docker run"
 	fi
 
+
+	if [[ -n $HZN_ICP_CA_CERT_PATH ]]; then
+		docker cp $HZN_ICP_CA_CERT_PATH $DOCKER_NAME:/usr/local/share/ca-certificates/icp.crt
+		docker exec $DOCKER_NAME update-ca-certificates
+		docker exec $DOCKER_NAME /root/anax.service restart
+	fi
+
```
bmpotter: i'm concerned about this approach, because if `docker run` returns before the container is fully up, the `docker exec` could fail. I think the better way to handle this is: if `$HZN_ICP_CA_CERT_PATH` is set then add another mount for it to the `docker run`, and then have `anax.service` (which runs inside the container) detect the existence of that file/dir, copy it to `/usr/local/share/ca-certificates/` and do `update-ca-certificates`. If it does this at the right time, it won't even require a 2nd restart of anax (because `anax.service` already modifies anax.json and restarts anax).

bmpotter: Also, `anax.json` has a field `Edge.CACertsPath`. How does that interact with the new variable in `/etc/default/horizon` ?

MaxMcAdam: They do not interact directly, certificates in the CACertsPath will still be trusted.

Issue #1014, opened 2019-06-13
Environment variables should override configuration files.


TheMosquito:
**From Bruce:**
I've changed my mind about this, and agree with glen: if `HZN_ORG_ID` is set, it should take precedence over it being set in a local project file (`horizon/hzn.json`). It is too confusing otherwise.

**From Glen:**
Environment variables should always have higher precedence than config files.

I have had difficulty debugging issues with this in my own use of Horizon, and I have had to help David Martin, and Joe Pearson to resolve these issues as well. Ivan's review of the documentation also ran into this issue. He said:

> 4.3. hzn exchange user list & hzn exchange node create -n $HZN_EXCHANGE_NODE_AUTH
> Did not work until I updated ./horizon/hzn.json with my own org id (as explained in > https://github.com/open-horizon/examples/blob/master/edge/services/helloworld/README.md#process-for-the-horizon-development-team-to-make-updates-to-the-helloworld-service )
> “HZN_ORG_ID”: “IBM”,
> To:
> “HZN_ORG_ID”: “bmtutor@us.ibm.com”,

So the above commands will not work as expected, unless you are outside of any directory that contains a horizon subdirectory containing hzn.json. He happened to run these commands inside an example directory so it did not work as expected.

Changing into a directory should not override configuration in your shell. This makes the act of changing directories have a significant side effect on the behavior of Horizon commands.
Also, intuitively, one would expect the environment variable to take precedence over static files, and CLI arguments to take precedence over environment variables. This precedence order is simply the order of closeness to the user: CLI (closest), shell environment (next layer), file system (and of course within the file system the is another similar order of precedence for the various supported configuration files).

jiportilla:
will also fix some typos on the error msgs in hzn.go like :    mmsObjectPublishPat := mmsObjectPublishCmd.Flag("pattern", "If you want the object to be deployed on nodes using a given pattern, specify it using this flag. This flag is optionla and can only be used with --type and --id. It is mutually exclusive with -m").Short('p').String()

jiportilla:
PR submitted

bmpotter:
Troy, please verify this and put a comment here about the results. The situation this comes up in the most is the when you are in 1 of the sample dirs and run an hzn cmd. If you don't have the `HZN_ORG_ID` env var set, it should default to the

PR #1100, opened 2019-07-25
issue #1014 - Env variables override config files

jiportilla:
override = false for all setenvvars

### DIFF
--- a/cli/cliconfig/config.go
+++ b/cli/cliconfig/config.go
@@ -341,7 +341,7 @@ func ReadJsonFileWithLocalConfig(filePath string) []byte {
 	metadata_vars := map[string]string{}
 	if useLocalConfig {
 		orig_env_vars = GetEnvVars()
-		hzn_vars, metadata_vars, err = SetEnvVarsFromConfigFile(localConfigFile, orig_env_vars, true)
+		hzn_vars, metadata_vars, err = SetEnvVarsFromConfigFile(localConfigFile, orig_env_vars, false)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


cli/cliconfig/config.go
```diff
@@ -417,7 +417,7 @@ func SetEnvVarsFromConfigFiles(project_dir string) error {
 			cliutils.Fatal(cliutils.CLI_GENERAL_ERROR, "Failed to get the absolute path for file %v. %v", configFile_project, err)
 		}
 		if configFile_project != configFile_pkg && configFile_project != configFile_user {
-			_, _, err = SetEnvVarsFromConfigFile(configFile_project, orig_env_vars, true)
+			_, _, err = SetEnvVarsFromConfigFile(configFile_project, orig_env_vars, false)
```
dabooz: I think like 344 in this same file has a call to SetEnvVarsFromConfigFile that should also be changed to pass false for the override parameter.

jiportilla:
I did not find a true flag on that line (344) but found one on 

if useLocalConfig {
		orig_env_vars = GetEnvVars()
		hzn_vars, metadata_vars, err = SetEnvVarsFromConfigFile(localConfigFile, orig_env_vars, true)
		if err != nil {
			cliutils.Fatal(cliutils.CLI_GENERAL_ERROR, "Failed to set the environment variables from the local configuration file %v for file %v. Error: %v", localConfigFile, filePath, err)
		}
	}

cli/cliconfig/config.go
jiportilla: updated override = false in all calls

PR #1101, opened 2019-07-26
'hzn exchange business listpolicy' shouldn't show an error when there…

MaxMcAdam:
… are no business policies in the exchange

### DIFF
--- a/cli/exchange/business.go
+++ b/cli/exchange/business.go
@@ -70,10 +70,9 @@ func BusinessListPolicy(org string, credToUse string, policy string, namesOnly b
 	if httpCode == 404 && policy != "" {
 		cliutils.Fatal(cliutils.NOT_FOUND, "Policy %s not found in org %s", policy, polOrg)
 	} else if httpCode == 404 {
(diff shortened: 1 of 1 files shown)


cli/exchange/business.go
```diff
@@ -70,10 +70,9 @@ func BusinessListPolicy(org string, credToUse string, policy string, namesOnly b
 	if httpCode == 404 && policy != "" {
 		cliutils.Fatal(cliutils.NOT_FOUND, "Policy %s not found in org %s", policy, polOrg)
 	} else if httpCode == 404 {
-		cliutils.Fatal(cliutils.NOT_FOUND, "Business policy for organization %s not found", polOrg)
-	}
-
-	if namesOnly && (policy == "" || policy == "*") {
+		policyNameList := []string{}
+		fmt.Println(policyNameList)
+	} else if namesOnly && (policy == "" || policy == "*") || httpCode == 404 {
```
linggao: no need to put `httpCode == 404` here because it has been handled above.

PR #1135, opened 2019-08-14
Issue 1106 - CLI messages globalized

linggao:
Please ignore the "Anax globalization" part when doing the review.


jiportilla:
check man on Pi
post-pone date format , let use UTC

PR #1151, opened 2019-08-20
Issue 935 - e2edev remote

njabbouribm:


test/Makefile
```diff
@@ -312,7 +323,10 @@ run-css: run-mongo-db
 		--name "$(DOCKER_CSS_CNAME)" \
 		-p 127.0.0.1:9443:9443 \
 		--network "$(DOCKER_TEST_NETWORK)" \
-		-e "HZN_EXCHANGE_URL=http://$(DOCKER_EXCH_CNAME):8080/v1" \
+		-e "HZN_EXCHANGE_URL=$(DOCKER_EXCH)" \
+		-e "CSS_URL=$(CSS_URL)" \
```
dabooz: I know we need the HZN_EXCHANGE_URL set, do we need the other 3 for the CSS container?

test/gov/agbot_del_loop.sh
```diff
@@ -4,24 +4,43 @@
...
-   DEL=$(curl -sS -X DELETE http://localhost:81/agreement/$AGID2)
+     echo "Deleting $AGID3"
+     DEL=$(curl -sS -X DELETE http://localhost:81/agreement/$AGID3)
 
-   echo "Deleting $AGID3"
-   DEL=$(curl -sS -X DELETE http://localhost:81/agreement/$AGID3)
+     echo "Deleting $AGID4"
+     DEL=$(curl -sS -X DELETE http://localhost:81/agreement/$AGID4)
+   else
+     AGID1=$(kubectl -n kube-system exec -ti $(kubectl -n kube-system get pods | grep -v 'agbot-db' | grep -m 1 'edge-computing-agbot' |  awk '{print $1}') -- sh -c "export ANAX_PORT=\$(cat /etc/horizon/anax.json | jq -r .AgreementBot.APIListen | awk -F: '{print \$NF}'); curl http://localhost:\$ANAX_PORT/agreement | jq -r '.agreements.active[0].current_agreement_id'")
```
dabooz: I thought this entire path was all going to be commented out until we did more work? I dont think this will work for anyone else who tries to run this remotely?

test/Makefile
njabbouribm: You are correct, we do not need the other 3 for the CSS container. I just ran a test to confirm.

test/gov/agbot_del_loop.sh
njabbouribm: It does not work, but also does not affect the run. It will give the user an unauthorized message, but that is the extent of the damage.

dabooz: OK, that's a fair point.

PR #1182, opened 2019-08-30
Issue935 - ICP support

njabbouribm:


test/docker/fs/etc/colonus/anax-combined.config.tmpl
```diff
@@ -10,6 +10,7 @@
         "PublicKeyPath": "/root/.colonus/mtn-publicKey.pem",
         "DefaultServiceRegistrationRAM": 1024,
         "ExchangeURL": "${EXCH_APP_HOST}/",
+        "CACertsPath": "/certs/css.crt",
```
dabooz: There is already a CACertsPath field in this file...and why arent you setting it in the remote config?

test/gov/gov-combined.sh
```diff
@@ -149,135 +149,260 @@ if [ "${EXCH_APP_HOST}" = "http://exchange-api:8080/v1" ]; then
 else
 
   cd /root
+  if [ ${CERT_LOC} -eq "1" ]; then
+    echo "Delete e2edev@somecomp.com..."
```
dabooz: I think you need to put both side of this if then else into a script that you can condition with a parameter. There's just too much duplication here now.

test/docker/fs/etc/colonus/anax-combined.config.tmpl
njabbouribm: Sorry the naming convention is wrong. I had originally named it remote due to the fact that it was being used for DEV. I need to change it from Remote to no-cert, because now the remote is used for anytime that does not require a cert and regular is for when it does require a cert. I had to include the CACertsPath, because I had a command that was failing even when using HZN_SSL_SKIP_VERIFY=1.

test/gov/gov-combined.sh
njabbouribm: I agree with your statement, I will work on something. I think the majority of the if statements are there to either add the --cert or remove it in the curl command.

PR #1192, opened 2019-09-04
Issue1096 - Agbots verify node agreements

dabooz:


producer/basic_protocol_handler.go
```diff
@@ -105,14 +105,49 @@ func (c *BasicProtocolHandler) VerifyAgreement(ag *persistence.EstablishedAgreem
...
+			}
+		}
+
+		return true, false, verify.AgreementId(), nil
+
+	} else {
+
+		// Not a known protocol extension message. The only protocol message that is not handled in this code path is the proposal
+		// so make sure it's not one of those, then we know if it's an unknown msg or not.
+		if _, err := c.agreementPH.ValidateProposal(msg.ProtocolMessage()); err == nil {
```
linggao: Just want to make sure that we do not validate it twice because ValidateProposal is called in agreement.go too.

PR #1193, opened 2019-09-05
Issue 783 - 'hzn register' handle node pattern defined in the exchange

linggao:
Did the pattern checking for both hzn and anax api because we would like to give proper messages to the user who are using `hzn register` in some special cases.

### DIFF
--- a/cli/cliutils/cliutils.go
+++ b/cli/cliutils/cliutils.go
@@ -322,9 +322,22 @@ func TrimOrg(org, id string) (string, string) {
 	} else if len(substrings) == 2 {
 		return substrings[0], substrings[1] // in this case the org the prepended to the id will override the org they may have specified thru the -o flag or env var
 	} else {
-		Fatal(CLI_INPUT_ERROR, i18n.GetMessagePrinter().Sprintf("the resource id can not contain more than 1 '/'"))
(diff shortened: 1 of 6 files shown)


cli/cliutils/cliutils.go
```diff
@@ -327,6 +327,19 @@ func TrimOrg(org, id string) (string, string) {
...
 
+// Add the given org to the id if the id does not already contain an org
+func AddOrg(org, id string) string {
+	substrings := strings.Split(id, "/")
+	if len(substrings) <= 1 { // this means id was empty, or did not contain '/'
+		return fmt.Sprintf("%v/%v", org, id)
+	} else if len(substrings) == 2 {
+		return id
+	} else {
+		Fatal(CLI_INPUT_ERROR, i18n.GetMessagePrinter().Sprintf("the resource id can not contain more than 1 '/'"))
```
dabooz: Why does the message say "the resource id can not" instead of "the id can not". What is the word resource referring to?

linggao: It is copied from the function TrimOrg() above the AddOrg().  TrimOrg is called in many places. The id passed in could be a pattern id, node id, service id. I guess the old code uses `resource id` as a general term. Shall I change "resource id" in both places to just "id"?

dabooz: So yes, I think we should remove the word resource from both places

linggao: Done.

PR #1194, opened 2019-09-05
Add e2edev test case for surfaced errors. Add workload and timestamp …

MaxMcAdam:
…to surface errors. Fix -l flag on 'hzn exchange node listerrors'

### DIFF
--- a/api/path_eventlog.go
+++ b/api/path_eventlog.go
@@ -32,23 +32,16 @@ func FindEventLogsForOutput(db *bolt.DB, all_logs bool, selections map[string][]
 	}
 }
 
-func FindSurfaceLogsForOutput(db *bolt.DB, msgPrinter *message.Printer) ([]persistence.EventLog, error) {
-	eventLogs := make([]persistence.EventLog, 0)
-
-	if surfaceLogs, err := persistence.FindSurfaceErrors(db); err != nil {
+func FindSurfaceLogsForOutput(db *bolt.DB, msgPrinter *message.Printer) ([]persistence.SurfaceError, error) {
+	outputLogs := make([]persistence.SurfaceError, 0)
+	surfaceLogs, err := persistence.FindSurfaceErrors(db)
+	if err != nil {
 		return nil, err
-	} else {
-		for _, log := range surfaceLogs {
-			if !log.Hidden {
-				recordSelector := []persistence.Selector{persistence.Selector{Op: "=", MatchValue: log.Record_id}}
-				recordSelectorMap := make(map[string][]persistence.Selector)
-				recordSelectorMap["record_id"] = recordSelector
-				fullEventlog, err := eventlog.GetEventLogs(db, true, recordSelectorMap, msgPrinter)
-				if err == nil && len(fullEventlog) > 0 {
-					eventLogs = append(eventLogs, fullEventlog[0])
-				}
-			}
+	}
+	for _, log := range surfaceLogs {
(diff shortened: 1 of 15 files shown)


test/gov/gov-combined.sh
```diff
@@ -500,6 +500,9 @@ then
         fi
     fi
 
+    ./verify_surfaced_error.sh
+    if [ $? -ne 0 ]; then echo "Verify surfaced error failure."; exit 1; fi
```
dabooz: I think this needs to be further down the script, around line 571

test/gov/verify_surfaced_error.sh
```diff
@@ -0,0 +1,96 @@
...
+  ERRS=$(curl -sSL http://localhost/eventlog/surface)
+  NUM_ERRS=$(echo ${ERRS} | jq -r '. | length')
+  sleep 5s
+  TIMEOUT = $TIMEOUT+1
+done
+
+echo -e "Found surfaced error $ERRS"
+
+# cpu service - needed by the hzn dev tests and the location top level service as a 3rd level dependency.
+VERS="1.2.2"
```
dabooz: I suspect this needs to be 1.2.4 in order to get the error to stop. And I think you need to add some checking that the errors have stopped.

test/gov/verify_surfaced_error.sh
```diff
@@ -0,0 +1,96 @@
+
+echo "Testing node error surfacing to exchange"
+
+IBM_ADMIN_AUTH="IBM/ibmadmin:ibmadminpw"
+KEY_TEST_DIR="/tmp/keytest"
+export HZN_EXCHANGE_URL="${EXCH_APP_HOST}"
+
```
dabooz: I need you to verify with Nadim that the 3 preceding lines are not going to cause a problem with remote testing.

api/path_eventlog.go
```diff
@@ -32,23 +32,11 @@ func FindEventLogsForOutput(db *bolt.DB, all_logs bool, selections map[string][]
...
 }
 
-func FindSurfaceLogsForOutput(db *bolt.DB, msgPrinter *message.Printer) ([]persistence.EventLog, error) {
-	eventLogs := make([]persistence.EventLog, 0)
-
-	if surfaceLogs, err := persistence.FindSurfaceErrors(db); err != nil {
+func FindSurfaceLogsForOutput(db *bolt.DB, msgPrinter *message.Printer) ([]persistence.SurfaceError, error) {
+	surfaceLogs := make([]persistence.SurfaceError, 0)
+	surfaceLogs, err := persistence.FindSurfaceErrors(db)
+	if err != nil {
```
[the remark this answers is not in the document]
MaxMcAdam: Added the filter back in

test/gov/verify_surfaced_error.sh
MaxMcAdam: Okayed by Nadim

MaxMcAdam:
Some things to note: 

- I added `hzn policy patch …` in the e2edev test so that the node will switch to the newer version of the service faster, otherwise it will sometimes timeout before the new version is deployed. 

- Set the config var in e2edev for how long an agreement has to execute before the relevant errors will be dismissed. 

- The recursive logic for finding all services deployed by an agreement is mostly copied from a similar function in the governance worker. 

- The changes in container/container.go are so that service errors include the workload org.

PR #1196, opened 2019-09-05
Issue 1156 - New Agbot API to display config file information

jiportilla:
curl localhost:8082/config | jq

### DIFF
--- a/agreementbot/api.go
+++ b/agreementbot/api.go
@@ -29,9 +29,10 @@ type API struct {
 	EC             *worker.BaseExchangeContext
 	em             *events.EventStateManager
 	shutdownError  string
+	configFile     string
 }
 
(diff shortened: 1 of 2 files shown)


agreementbot/api.go
```diff
@@ -641,8 +663,23 @@ func NewHorizonAgbot(id string, org string) *HorizonAgbot {
 	}
 }
 
-func getAgbotInfo(config *config.HorizonConfig) {
+type HorizonAgbotConfig struct {
+	InMemoryConfig   config.AGConfig `json:"In_Memory_Config"`
```
dabooz: Minor tweak. We try to use camel case, so the json field names should be InMemoryConfig and FileSystemConfig.

PR #1239, opened 2019-09-13
Only show mms object detail if -d is specified

MaxMcAdam:


MaxMcAdam:
Getting only the specified object from css does not return the destination status array that getting all the objects of that type does. That's why I did it in this way with the loop, not a more efficient way of getting just the specified object.

PR #1273, opened 2019-10-04
Issue 694 - e2edev testing using hzn command to register.

linggao:


test/gov/hzn_reg.sh
```diff
@@ -1,52 +1,147 @@
...
-cd /root/hzn/helloms
-hzn dev microservice publish -vI -k $KEY_TEST_DIR/*private.key
-if [ $? -ne 0 ]
-then
-    echo -e "hzn dev microservice publish failed."
-    exit 1
+# register
+if [ -n "$PATTERN" ]; then 
+	# register pattern sall
+	echo -e "  hzn register -u $E2EDEV_ADMIN_AUTH -n an12345:abcdefg -o e2edev@somecomp.com -f /tmp/reg_userinput.json -p e2edev@somecomp.com/$PATTERN"
```
dabooz: We need some additional tests that omit the -o flag, where the pattern is specified without the org, where the auth is specified without the org, where HZN_ORG_ID is used, etc etc. Also we need to test for the node already registered case, and the already unregistered case. Since it takes so long for all the agreements in the sall caseto be created, maybe we use the sns pattern (and the ns policy) to test this because that will have fewer workloads to start.

PR #1274, opened 2019-10-07
Issue844 - tail -F

njabbouribm:


cli/service/service.go
```diff
@@ -76,6 +81,59 @@ func List() {
...
+				continue
+			}
+			new_file_size := fi.Size()
+			// if new_file_size is smaller than previous measured size,
+			// log was logrotated
+			if new_file_size >= file_size {
+				file_size = new_file_size
+			} else {
+				// new file detected
+				fmt.Printf("\nNew File Detected\n")
```
dabooz: Please delete this

cli/service/service.go
```diff
@@ -76,6 +81,59 @@ func List() {
...
+			cliutils.Fatal(cliutils.NOT_FOUND, msgPrinter.Sprintf("/var/log/syslog could not get stats", err))
+		}
+		file_size := fi.Size()
+		defer file.Close()
+		reader := bufio.NewReader(file)
+		for {
+			// open file again to re-check stats in case it was logrotated
+			new_file, err := os.Open("/var/log/syslog")
+			if err != nil {
+				continue
```
dabooz: Why continue? If you want to retry the open because of log rotate then you should probably also sleep for a second.  Same question on the Stat call below.

PR #1276, opened 2019-10-15
Issue 844 - hzn service log

njabbouribm:
**Still need to implement:**

"As a related item, because this will simplify the testing of the helloworld service, i'd like you to add a test target to the Makefile that hzn dev service new ... generates. See https://github.com/open-horizon/examples/blob/master/edge/services/helloworld/Makefile"

- and -

"On mac, you must do:

docker logs -f $(docker ps -q --filter name=<service-name>)"

### DIFF
--- a/cli/service/service.go
+++ b/cli/service/service.go
@@ -1,6 +1,7 @@
 package service
 
 import (
+	"bufio"
 	"encoding/json"
 	"fmt"
 	"github.com/open-horizon/anax/api"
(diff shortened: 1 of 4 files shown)


cli/service/service.go
```diff
@@ -76,6 +82,84 @@ func List() {
 	fmt.Printf("%s\n", jsonBytes)
 }
 
+func Log(ServiceName string) {
+	msgPrinter := i18n.GetMessagePrinter()
+	ref_url := "\"ref_url\": \"" + ServiceName + "\""
+	re := regexp.MustCompile(`instance_id\": \"(.*?)\"`)
+	var inputs api.AllServices
+	cliutils.HorizonGet("service", []int{200}, &inputs, false)
+	_, err := cliutils.DisplayAsJson(inputs.Instances["active"])
```
dabooz: Why are you processing the response as JSON? You should be using the returned object directly and then iterating through the structure to find the service you're looking for.

PR #1280, opened 2019-10-18
Issue 1202 - anax detect pattern changes after started with a pattern

linggao:


exchangesync/node.go
```diff
@@ -0,0 +1,69 @@
...
+// Get the node from the exchange and save it
+func SyncNodeWithExchange(db *bolt.DB, pDevice *persistence.ExchangeDevice, getDevice exchange.DeviceHandler) (*exchange.Device, error) {
+
+	glog.V(4).Infof("Checking the node changes.")
+
+	nodeUpdateLock.Lock()
+	defer nodeUpdateLock.Unlock()
+
+	// get the node user input from the exchange
+	exchDevice, err := getDevice(fmt.Sprintf("%v/%v", pDevice.Org, pDevice.Id), pDevice.Token)
```
dabooz: If you need this lock, should you really be holding it across a network call that could retry forever?

config/config.go
```diff
@@ -54,10 +54,11 @@ type Config struct {
 	MultipleAnaxInstances            bool      // multiple anax instances running on the same machine
 	DefaultServiceRetryCount         int       // the default service retry count if retries are not specified by the policy file. The default value is 2.
 	DefaultServiceRetryDuration      uint64    // the default retry duration in seconds. The next retry cycle occurs after the duration. The default value is 600
-	ServiceConfigStateCheckIntervalS int       // the service configuration state check interval. The default is 30 seconds.
+	ServiceConfigStateCheckIntervalS int       // deprecated. The service configuration state check interval. The default is 30 seconds.
 	DefaultNodePolicyFile            string    // the default node policy file name.
+	NodeCheckIntervalS               int       // the node check interval. The default is 15 seconds.
```
dabooz: Doug Larson is playing with these interval settings, so you need to tell him that you just consolidated some of the pollers and put them into 1 config switch.

config/config.go
```diff
@@ -54,10 +54,11 @@ type Config struct {
 	MultipleAnaxInstances            bool      // multiple anax instances running on the same machine
 	DefaultServiceRetryCount         int       // the default service retry count if retries are not specified by the policy file. The default value is 2.
 	DefaultServiceRetryDuration      uint64    // the default retry duration in seconds. The next retry cycle occurs after the duration. The default value is 600
-	ServiceConfigStateCheckIntervalS int       // the service configuration state check interval. The default is 30 seconds.
+	ServiceConfigStateCheckIntervalS int       // deprecated. The service configuration state check interval. The default is 30 seconds.
```
dabooz: It doesnt look like there is any code which references this field, so I think you should delete it.

config/config.go
```diff
@@ -54,10 +54,11 @@ type Config struct {
...
 	DefaultNodePolicyFile            string    // the default node policy file name.
+	NodeCheckIntervalS               int       // the node check interval. The default is 15 seconds.
 	NodePolicyCheckIntervalS         int       // the node policy check interval. The default is 15 seconds.
-	NodeUserInputCheckIntervalS      int       // the node user input check interval. The default is 15 seconds.
+	NodeUserInputCheckIntervalS      int       // deprecated. The node user input check interval. The default is 15 seconds.
```
dabooz: It doesnt look like there is any code which references this field, so I think you should delete it.

governance/governance.go
```diff
@@ -44,6 +44,7 @@ const CONTAINER_GOVERNOR = "ContainerGovernor"
 const MICROSERVICE_GOVERNOR = "MicroserviceGovernor"
 const BC_GOVERNOR = "BlockchainGovernor"
 const SERVICE_CONFIGSTATE_GOVERNOR = "ServiceConfigStateGovernor"
```
dabooz: I think this constant is not used any more

governance/node.go
```diff
@@ -0,0 +1,233 @@
...
+				glog.Errorf(logString(fmt.Sprintf("error persisting unconfiguring on node object: %v", err)))
+				return
+			}
+			// set the node shutdown message
+			w.Messages() <- events.NewNodeShutdownMessage(events.START_UNCONFIGURE, false, false)
+		}
+	} else {
+		// the device is up again and rereg the device with the new pattern
+		if err := w.changeNodePattern(pDevice, new_pattern); err != nil {
+			glog.Errorf(logString(fmt.Sprintf("error while re-registeting nnode with new pattern %v. %v", new_pattern, err)))
```
dabooz: 2 typos on this line

governance/node.go
```diff
@@ -0,0 +1,233 @@
...
+			if err != nil {
+				glog.Errorf(logString(fmt.Sprintf("error persisting unconfiguring on node object: %v", err)))
+				return
+			}
+			// set the node shutdown message
+			w.Messages() <- events.NewNodeShutdownMessage(events.START_UNCONFIGURE, false, false)
+		}
+	} else {
+		// the device is up again and rereg the device with the new pattern
+		if err := w.changeNodePattern(pDevice, new_pattern); err != nil {
```
dabooz: What state is the device in if this error occurs? Is it unregistered? How does the user recover?

governance/shutdown.go
```diff
@@ -22,7 +22,6 @@ import (
 // There are other workers responsible for other functions, which will also so some cleanup when the Node Shutdown Message
 // arrives. For example, the node heartbeat function is stopped by the Agreement worker.
 func (w *GovernanceWorker) nodeShutdown(cmd *NodeShutdownCommand) {
-	glog.V(3).Infof(logString(fmt.Sprintf("begin node shutdown process.")))
```
dabooz: It is really important for this debug message to remain here. Moving it down isnt helpful. I know we will get 2 shutdown begin msgs but they are slightly different, I think that's good because its clear what is happening.

agreement/node.go
```diff
@@ -0,0 +1,367 @@
...
+			eventlog.LogDatabaseEvent(w.db, persistence.SEVERITY_ERROR,
+				persistence.NewMessageMeta(EL_AG_UNABLE_WRITE_NODE_EXCH_PATTERN_TO_DB, exchDevice.Pattern, err.Error()),
+				persistence.EC_DATABASE_ERROR)
+			return
+		} else {
+			glog.Infof(logString(fmt.Sprintf("Node pattern changed on the exchange from %v to %v. Will re-register the node.", w.devicePattern, exchDevice.Pattern)))
+			eventlog.LogDatabaseEvent(w.db, persistence.SEVERITY_INFO,
+				persistence.NewMessageMeta(EL_AG_NODE_PATTERN_CHANGED, w.devicePattern, exchDevice.Pattern),
+				persistence.EC_NODE_PATTERN_CHANGED)
+			w.Messages() <- events.NewNodePatternMessage(events.NODE_PATTERN_CHANGE_SHUTDOWN, exchDevice.Pattern)
```
dabooz: You can't send the shutdown event until you verify that the new pattern can run on this node, i.e. that all the services in the new pattern have been configured on the device. If the new pattern cant run here then we should log that in the error log and perhaps this is one of the errors we should surface to the exchange.

exchangesync/node.go
linggao: Good catch. Fixed.

config/config.go
linggao: done

config/config.go
linggao: Done

config/config.go
linggao: Done

governance/governance.go
linggao: Done

governance/node.go
linggao: oops. fixed.

governance/shutdown.go
linggao: ok, ok, moved it up there.

governance/node.go
linggao: configured. First the pattern has been verified before shutdown.

linggao: If anything is wrong, user can set the pattern back on the exchange.

agreement/node.go
linggao: Added code to do verification.

governance/shutdown.go
```diff
@@ -118,8 +128,72 @@ func (w *GovernanceWorker) nodeShutdown(cmd *NodeShutdownCommand) {
...
+
+	// Remove the node’s messaging public key from the node’s exchange resource and delete the node’s message key pair from the filesystem.
+	if err := w.patchNodeKey(); err != nil {
+		w.continueWithError(logString(err.Error()))
+	}
+
+	// Add the new pattern name back to the node for the pattern change case, It is safe to add it back because the public keys are cleared
+	// from the node
+	if err := w.patchNodePattern(new_pattern); err != nil {
+		w.completedWithError(logString(err.Error()))
```
dabooz: I think this should be continueWithError

linggao: changed to continueWithError

PR #1282, opened 2019-10-18
Issue 1254 - Continue unregistration if exchange api call fails with non-transport…

MaxMcAdam:
… error

Unrelated formatting changes from 'make format'


governance/shutdown.go
```diff
@@ -50,8 +49,7 @@ func (w *GovernanceWorker) nodeShutdown(cmd *NodeShutdownCommand) {
 
 	// Remove the node’s messaging public key from the node’s exchange resource and delete the node’s message key pair from the filesystem.
 	if err := w.patchNodeKey(); err != nil {
-		w.completedWithError(logString(err.Error()))
-		return
+		w.continueWithError(logString(err.Error()))
```
dabooz: Line 80 also needs this same update.

PR #1288, opened 2019-10-23
Issue844 - Service Log Linux

njabbouribm:


cli/hzn.go
```diff
@@ -374,6 +374,9 @@ Environment Variables:
 	userinputRemoveForce := userinputRemoveCmd.Flag("force", msgPrinter.Sprintf("Skip the 'Are you sure?' prompt.")).Short('f').Bool()
 
 	serviceCmd := app.Command("service", msgPrinter.Sprintf("List or manage the services that are currently registered on this Horizon edge node."))
+	serviceLogCmd := serviceCmd.Command("log", msgPrinter.Sprintf("Show the container logs for a service."))
+	logServiceName := serviceLogCmd.Arg("service", msgPrinter.Sprintf("Service name. Use the url field of a service definition. Similar to tail behavior")).String()
```
dabooz: This is not the command structure I wanted. I think we want the service name to be required and the -f to be an optional flag that enables log rotation and continuous output.
logServiceName := serviceLogCmd.Arg("service", msgPrinter.Sprintf("The name of the service whose log records should be displayed. The service name is the same as the url field of a service definition. Displays log records similar to tail behavior and returns .")).Required().String()
logServiceNameTail := serviceLogCmd.Flag("service", msgPrinter.Sprintf("Continuously polls the service's logs to display the most recent records, similar to tail -F behavior.")).Short('f').Bool()

cli/service/service.go
```diff
@@ -76,6 +81,103 @@ func List() {
...
+		if strings.Contains(id.SpecRef, RefURL) {
+			instance_id = id.InstanceId
+			id_found = 1
+			break
+		}
+	}
+	if id_found == 0 {
+		cliutils.Fatal(cliutils.CLI_INPUT_ERROR, i18n.GetMessagePrinter().Sprintf("Service %v is not running on the node.", RefURL))
+	}
+	fmt.Printf("Displaying syslog messages for service %v with service id %v. Use ctrl-C to terminate this command.\n", RefURL, instance_id)
```
dabooz: RefURL is not the right variable to use here. I think it should be the id.SpecRef value.

cli/service/service.go
```diff
@@ -76,6 +81,103 @@ func List() {
 	fmt.Printf("%s\n", jsonBytes)
 }
 
+func Log(ServiceName string, ServiceNameTail string) {
+	msgPrinter := i18n.GetMessagePrinter()
+	RefURL := ServiceName
```
dabooz: Upper case has special meaning in go, it is used for function names that are visible outside of a package, and it is used for constants. Upper case works fine, this is a convention, not a language rule. Variable names should begin with lower case.

PR #1317, opened 2019-11-08
Issue 844 - Mac component

njabbouribm:


cli/service/service.go
```diff
@@ -110,6 +112,38 @@ func Log(serviceName string, tailing bool) {
 	if !serviceFound {
 		cliutils.Fatal(cliutils.CLI_INPUT_ERROR, msgPrinter.Sprintf("Service %v is not running on the node.", refUrl))
 	}
+	if runtime.GOOS == "darwin" {
```
dabooz: The code looks good. But I'd like you to do 2 things to clean it up a bit:
1) Move the mac and linux specific code into their own functions, and just call them from Log() function.
2) Add more comments to the mac code. Answer questions like; (a) when does the go func() end/terminate? (b) explain why you have to do the pipe and stderr/stdout setup that you're doing before the command is invoked

PR #1328, opened 2019-11-18
Issue 1327 - Change default port to 8510

dabooz:


config/config.go
```diff
@@ -172,6 +170,23 @@ func enrichFromEnvvars(config *HorizonConfig) error {
 		config.Edge.ExchangeMessageDynamicPoll = false
 	}
 
+	if apiPort := os.Getenv(AnaxAPIPort); apiPort != "" {
+		if config.Edge.APIListen != "" {
+			listen := strings.Split(config.Edge.APIListen, ":")
+			if len(listen) == 2 {
+				config.Edge.APIListen = fmt.Sprintf("%v:%v", listen[0], apiPort)
+			} else {
+				config.Edge.APIListen = fmt.Sprintf("127.0.0.1:%v", apiPort)
```
linggao: When len(listen) == 1, should it also be `fmt.Sprintf("%v:%v", listen[0], apiPort)`?

config/config.go
```diff
@@ -172,6 +170,23 @@ func enrichFromEnvvars(config *HorizonConfig) error {
...
+			}
+		} else {
+			config.Edge.APIListen = fmt.Sprintf("127.0.0.1:%v", apiPort)
+		}
+	} else {
+		if config.Edge.APIListen == "" {
+			config.Edge.APIListen = fmt.Sprintf("127.0.0.1:%v", AnaxAPIPortDefault)
+		}
```
linggao: how about the case when config.Edge.APIListen != "" but len(listen) ==1?  It should be `fmt.Sprintf("%v:%v", listen[0], AnaxAPIPortDefault)`

test/gov/agbot_del_loop.sh
```diff
@@ -52,7 +52,7 @@ do
       sleep 180
 
       echo -e "Current workload usages\n"
-      curl -sS http://localhost:81/workloadusage | jq -r '.'
+      curl -sS http://localhost:8082/workloadusage | jq -r '.'
```
linggao: Should it be 8511?

config/config.go
linggao: Please also change  the 2 lines here https://github.com/open-horizon/anax/blob/master/anax-in-container/horizon-container#L185-L188. Change 80 to 8510

config/config.go
dabooz: I am making the assumption that if len(listen) !=2 then it's an error because the format of the field is host:port, therefore len(listen)==1 means they only specified the host but not the port, that's why I set the host to loopback. How strongly do you feel about which way this should go?

config/config.go
dabooz: Will update horizon-container.
For the Edge.APIListen change, see above, not sure how it should work because we've never validated that field before, it's hard for the user to change and we set it to a default value.

test/gov/agbot_del_loop.sh
dabooz: No, 8082 is correct. This is the port for the agbot. It's the same port which is used earlier in the script.

config/config.go
linggao: I can go either way. Since you allow overwrite the port in the /etc/default/horizon, I was thinking that user could just put the ip in the APIListen and let the env variable decide the port number.

dabooz: ok, that would be a change to the currently defined format of the field. I need to think about it again.

anax-in-container/horizon-container
```diff
@@ -182,10 +182,10 @@ start() {
 	echo "Starting the Horizon agent container openhorizon/amd64_anax:$dockerTag..."
 	# Note: docker will automatically create the ${DOCKER_NAME}_var and ${DOCKER_NAME}_etc volumes if necessary
 	if isMacos; then
-		docker run -d -t --name $DOCKER_NAME --privileged -p 127.0.0.1:$HORIZON_AGENT_PORT:80 -e ANAX_DOCKER_ENDPOINT=tcp://host.docker.internal:$SOCAT_LISTEN_PORT -e HOST_OS=mac -e DOCKER_NAME=${DOCKER_NAME} $defaultFileMountArg $icpCertMount -v ${DOCKER_NAME}_var:/var/horizon/ -v ${DOCKER_NAME}_etc:/etc/horizon/ -v ${fssHostSharePath}:/var/tmp/horizon/${DOCKER_NAME} openhorizon/amd64_anax:$dockerTag
+		docker run -d -t --name $DOCKER_NAME --privileged -p 127.0.0.1:$HORIZON_AGENT_PORT:$HZN_AGENT_PORT -e ANAX_DOCKER_ENDPOINT=tcp://host.docker.internal:$SOCAT_LISTEN_PORT -e HOST_OS=mac -e DOCKER_NAME=${DOCKER_NAME} $defaultFileMountArg $icpCertMount -v ${DOCKER_NAME}_var:/var/horizon/ -v ${DOCKER_NAME}_etc:/etc/horizon/ -v ${fssHostSharePath}:/var/tmp/horizon/${DOCKER_NAME} openhorizon/amd64_anax:$dockerTag
```
linggao: Is will be safe to check if HZN_AGENT_PORT is defined or not, if not, make the  port map to 8510.

PR #1330, opened 2019-11-18
Issue 1081 hzn exchange node update verify userinput

LiilyZhang:
1. service does not exist
Screen Shot 2019-12-01 at 19 53 24
Screen Shot 2019-12-01 at 19 55 29
2. the variables are defined in the service, otherwise return a warning message
Screen Shot 2019-12-01 at 20 01 47

3. variables are correct type
Screen Shot 2019-12-01 at 20 01 06

4. the node is associated with a pattern `sall`, the service should be defined as top level service or as dependent service
Screen Shot 2019-12-01 at 19 59 54

5. 
Screen Shot 2019-12-01 at 20 15 28
the userinputs need to be set in given userinput.json if no default value given in service userinput.

### DIFF
--- a/cli/exchange/node.go
+++ b/cli/exchange/node.go
@@ -2,24 +2,26 @@ package exchange
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
 	"net/http"
 	"os"
 
 	"github.com/open-horizon/anax/cli/cliconfig"
 	"github.com/open-horizon/anax/cli/cliutils"
+	"github.com/open-horizon/anax/cutil"
 	"github.com/open-horizon/anax/exchange"
 	"github.com/open-horizon/anax/externalpolicy"
 	_ "github.com/open-horizon/anax/externalpolicy/text_language"
 	"github.com/open-horizon/anax/i18n"
 	"github.com/open-horizon/anax/persistence"
 	"github.com/open-horizon/anax/policy"
+	"github.com/open-horizon/anax/semanticversion"
 )
 
-// We only care about handling the node names, so the rest is left as interface{} and will be passed from the exchange to the display
 type ExchangeNodes struct {
-	LastIndex int                    `json:"lastIndex"`
-	Nodes     map[string]interface{} `json:"nodes"`
+	LastIndex int                        `json:"lastIndex"`
+	Nodes     map[string]exchange.Device `json:"nodes"`
 }
 
 func NodeList(org string, credToUse string, node string, namesOnly bool) {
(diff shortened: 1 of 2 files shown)


cli/exchange/node.go
```diff
@@ -155,6 +172,24 @@ func NodeUpdate(org string, credToUse string, node string, filePath string) {
 			if err := json.Unmarshal(bytes, &ui); err != nil {
 				cliutils.Fatal(cliutils.JSON_PARSING_ERROR, msgPrinter.Sprintf("failed to unmarshal attribute input %s: %v", v, err))
 			} else {
+				// validate userInput
+				if len(ui) != 0 {
+					msgPrinter.Printf("Start validating userinputs ....")
```
dabooz: I dont think you need the Start validating and validation is Done messages.

cli/exchange/node.go
```diff
@@ -465,3 +500,292 @@ func NodeListErrors(org string, credToUse string, node string, long bool) {
...
+func ValidateUserInput(org string, credToUse string, userInput policy.UserInput, exNode ExchangeNodes) (bool, string, error) {
+	msgPrinter := i18n.GetMessagePrinter()
+	warningMessage := ""
+
+	serviceOrg := userInput.ServiceOrgid
+	serviceUrl := userInput.ServiceUrl
+	serviceVersionRange := userInput.ServiceVersionRange
+	serviceArch := userInput.ServiceArch
+
+	msgPrinter.Printf("Start validating userinput %v, %v, %v, %v", serviceOrg, serviceUrl, serviceVersionRange, serviceArch)
```
dabooz: For this message just say "Validating userinput %v..", get rid of the "Start"

cli/exchange/node.go
```diff
@@ -465,3 +500,292 @@ func NodeListErrors(org string, credToUse string, node string, long bool) {
...
+	var sdef *exchange.ServiceDefinition
+
+	searchVersion, err := getSearchVersion(serviceVersion)
+	if err != nil {
+		return false, warningMessage, err
+	}
+
+	sdef, err = getService(cliutils.OrgAndCreds(org, credToUse), serviceOrg, serviceUrl, serviceArch, serviceVersion, searchVersion)
+	if sdef == nil {
+		errorString = msgPrinter.Sprintf("Service does not exist for org: %v, url: %v, version: %v, arch: %v, get error: %v \n", serviceOrg, serviceUrl, serviceVersion, serviceArch, err)
```
dabooz: If the org in the userinput is wrong, the errorString duplicates the text that's in the err variable. Either shorten the message in err, or just return the text in err. Here is the error message, see how the service does not exist is duplicated.
Error: Unable to validate exchange node userInput, error: Service does not exist for org: othere2edev@somecomp.com, url: https://bluehorizon.network/services/testservice, version: [0.0.0,INFINITY), arch: amd64, get error: Service does not exist for org: othere2edev@somecomp.com, url: https://bluehorizon.network/services/testservice, version: [0.0.0,INFINITY), arch: amd64

However, if the version range is too high, the err variable contains nil. You need to make the error reporting more consistent. Here's the error for this one:
Error: Unable to validate exchange node userInput, error: Service does not exist for org: e2edev@somecomp.com, url: https://bluehorizon.network/services/testservice, version: [5.0.0,INFINITY), arch: amd64, get error: <nil>

cli/exchange/node.go
```diff
@@ -465,3 +500,292 @@ func NodeListErrors(org string, credToUse string, node string, long bool) {
...
+
+		errorString = msgPrinter.Sprintf("Service %v %v %v %v in userInput is not defined in pattern or dependent services", serviceOrg, serviceUrl, serviceArch, vExp.Get_expression())
+		return false, warningMessage, errors.New(errorString)
+
+	}
+	return true, warningMessage, nil
+
+}
+
+func checkDependentServices(creds string, serviceOrg string, serviceArch string, serviceUrl string, serviceVersionExp *semanticversion.Version_Expression, servToCheckDependentService ServiceReference, version string, versionRange string) bool {
```
dabooz: Not sure where the problem is. If a node is using a pattern, and the top level service requires another service, and the required service has userinputs without default values, then we need to catch that case and issue an error.

There is also a related problem that is even more simple to create. If the top level service has a userinput with no default value AND that userinput is not set in the input body, that should also be flagged as an error.

cli/exchange/node.go
```diff
@@ -465,3 +480,341 @@ func NodeListErrors(org string, credToUse string, node string, long bool) {
...
+	sdef, err = getService(cliutils.OrgAndCreds(org, credToUse), serviceOrg, serviceUrl, serviceArch, serviceVersion, searchVersion)
+	if sdef == nil || err != nil {
+		return false, warningMessage, err
+	}
+
+	// compare ServiceDefinition.UserInput(array) with userInput.Inputs(array)
+	var compareResult bool
+	compareResult, err = compareUserInput(sdef.UserInputs, userInput.Inputs)
+	if !compareResult {
+		errorString = msgPrinter.Sprintf("Failed to compare given userInput with service userInput, error: %v", err)
```
dabooz: Please change the wording of the message to:
errorString = msgPrinter.Sprintf("Failed comparing userInput with service's userInput, error: %v", err)

cli/exchange/node.go
```diff
@@ -465,3 +480,341 @@ func NodeListErrors(org string, credToUse string, node string, long bool) {
...
+
+	// if service has a userinput with no default value AND userInput is not set in input body, return err
+	for _, serviceInput = range sdefUserInputs {
+		serviceInputName = serviceInput.Name
+		serviceUserInputsMap[serviceInputName] = serviceInput
+
+		if serviceInput.DefaultValue == "" {
+			valueDefined := sliceContains(cmdUserInputNames, serviceInputName)
+			if !valueDefined {
+				errorString = msgPrinter.Sprintf("%v need to be defined in userInput", serviceInputName)
```
dabooz: Please change error message to:
errorString = msgPrinter.Sprintf("%v needs to be set in userInput because it has no default value.", serviceInputName)

cli/exchange/node.go
```diff
@@ -465,3 +480,341 @@ func NodeListErrors(org string, credToUse string, node string, long bool) {
...
+		policyInputName = policyInput.Name
+		policyInputValue = policyInput.Value
+
+		serviceInput, ok = serviceUserInputsMap[policyInputName]
+		if !ok {
+			// give back a warning
+			inputNameNotDefinedInService = append(inputNameNotDefinedInService, policyInputName)
+		} else {
+			if err := cutil.VerifyWorkloadVarTypes(policyInputValue, serviceInput.Type); err != nil {
+				errorString = msgPrinter.Sprintf("Error validating type in user input %v. Error: %v", policyInputName, err)
```
dabooz: Please change this error message to:
errorString = msgPrinter.Sprintf("Error validating the type of userInput variable %v, error: %v", policyInputName, err)

cli/exchange/node.go
dabooz: The second problem seems to be fixed, but the first problem is not. A node with a pattern, where the pattern's top level service requires another service that has userinputs but no default values, is not flagged as in error.

cli/exchange/node.go
```diff
@@ -465,3 +480,341 @@ func NodeListErrors(org string, credToUse string, node string, long bool) {
...
+	nodePatternString := ""
+	for _, node := range nodes {
+		nodePatternString = node.Pattern
+	}
+
+	// Need to check:
+	// 1) if service is top level service
+	// 2) or recursively check if service defined in dependent service
+	// 3) if service range of the dependent service has intersection with versionRange of service from userInput
+	// 4) the required service has userinputs should have default values
```
dabooz: A node with a pattern, where the pattern's top level service requires another service that has userinputs but no default values, is not flagged as in error. Case 4 here does not appear to work correctly.

cli/exchange/node.go
```diff
@@ -465,3 +480,341 @@ func NodeListErrors(org string, credToUse string, node string, long bool) {
...
+		}
+
+		// recursive call
+		findMatchedService, _, _ := checkDependentServices(creds, serviceOrg, serviceArch, serviceUrl, serviceVersionExp, depService, "", dependentService.VersionRange, false)
+		if findMatchedService {
+			return true, true, nil
+		}
+	}
+
+	return false, true, errors.New(msgPrinter.Sprintf("Service %v, %v, %v, %v, %v does not define as dependent service", serviceOrg, serviceArch, serviceUrl, serviceVersionExp.Get_expression()))
```
dabooz: Please change the wording of this message:
...does not define a dependent service...
Also, I'm not sure this error is ever used. I think it gets ignored. If so, maybe be better to just return nil for it.

cli/exchange/node.go
```diff
@@ -465,3 +535,474 @@ func NodeListErrors(org string, credToUse string, node string, long bool) {
 		fmt.Printf("%s\n", jsonBytes)
 	}
 }
+
+func ValidateUserInput(org string, credToUse string, userInput policy.UserInput, topLevelServices []ServiceReference, userInputsMap map[string]policy.UserInput, patternUserInputsMap map[string]policy.UserInput, userInputWithEmptyDefaultValueMap map[string][]exchange.UserInput) (bool, string, error) {
```
dabooz: Somewhere we need checks to verify that the input userinput structure is valid. If you pass in a valid JSON struct that doesnt have all the right fields, you will get wierd errors:
for input {"userInput":[{"foo":"bar"}]}
you get:
Start validating userinput , , ,
Error: bad HTTP code 405 from GET http://exchange-api:8080/v1/orgs//services?url=&arch=, output:

LiilyZhang: Will return `Error: ServiceOrgid and ServiceUrl need to be set in userinput` in new pr

cli/exchange/node.go
```diff
@@ -465,3 +535,474 @@ func NodeListErrors(org string, credToUse string, node string, long bool) {
...
+			var topLevelServiceVersion ServiceChoice
+			if serviceOrg == topLevelService.ServiceOrg && serviceArch == topLevelService.ServiceArch && serviceUrl == topLevelService.ServiceURL {
+				for _, topLevelServiceVersion = range topLevelService.ServiceVersions {
+					if withinRange, _ := vExp.Is_within_range(topLevelServiceVersion.Version); withinRange == true {
+						sdef, err = getService(cliutils.OrgAndCreds(org, credToUse), topLevelService.ServiceOrg, topLevelService.ServiceURL, topLevelService.ServiceArch, topLevelServiceVersion.Version, "")
+						if sdef == nil || err != nil {
+							return false, warningMessage, errors.New(msgPrinter.Sprintf("toplevel service does not exist %v, %v, %v, %v", topLevelService.ServiceOrg, topLevelService.ServiceURL, topLevelService.ServiceArch, topLevelServiceVersion.Version))
+						}
+						foundService = true
+						//return true, warningMessage, nil
```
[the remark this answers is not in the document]
LiilyZhang: I think it needs to continue on the for loop even though we find service defined in toplevel. It needs to continue check and make sure any var with no default values (from service + pattern) are set in userinput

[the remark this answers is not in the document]
dabooz: ok

Issue #1340, opened 2019-11-21
agent install script uses an old autocomplete script location


ig0r:
Problem:

After the agent is installed on a node, a new shell session gives an error:

```bash
-bash: /usr/local/share/horizon/hzn_bash_autocomplete.sh: No such file or directory
```

The updated autocomplete script location should be used.

PR #1341, opened 2019-11-21
Small improvements for the agent install script

ig0r:
- a fix for autocomplete issue #1340
- removed a double negative in the message


agent-install/agent-install.sh
```diff
@@ -706,7 +706,7 @@ function process_node(){
 			# with a pattern currently
 			log_notify "The node currently has workload(s) (check them with hzn agreement list)"
 			if [[ -z "$HZN_EXCHANGE_PATTERN" ]] && [[ -z "$HZN_NODE_POLICY" ]]; then
-				log_info "Neither a pattern nor node policy has not been specified"
+				log_info "Neither a pattern nor node policy has been specified"
```
dabooz: Actually, you can register without pattern and without policy. The simple 'hzn register' command will cause your node to register with just the default builtin policy values.  Both -p and --policy are optional, but mutually exclusive.

PR #1356, opened 2019-12-09
Issue 1355 - correctly generate default node id

dabooz:


cli/register/register.go
```diff
@@ -137,7 +137,28 @@ func DoIt(org, pattern, nodeIdTok, userPw, email, inputFile string, nodeOrgFromF
 		nodeId = *horDevice.Id
 
 		if nodeId == "" {
-			cliutils.Fatal(cliutils.CLI_INPUT_ERROR, msgPrinter.Sprintf("Please specify the node id and token using -n flag or HZN_EXCHANGE_NODE_AUTH environment variable."))
+			// Generate a node id using the machine's serial number, if available.
+			var msErr error
+			nodeId, msErr = cutil.GetMachineSerial("")
+			if msErr != nil {
+				cliutils.Verbose(msgPrinter.Sprintf("Unable to read machine serial number, error: %v. Continuing device registration.", msErr))
```
linggao: Should we ignore the error and generate a random node id anyway? For the mac case, the cutil.GetMachineSerial always return error.

linggao: And I also think line 133 to 137 should be 
if horDevice.Id != nil && *horDevice.Id != "" {
     nodeId = *horDevice.Id
}

dabooz: If we ignore the error, the security by design scans will flag the fact that the error is being ignored which will cause us more work to respond to the false positive finding. And more importantly, the code is actually ignoring the error, it's just putting out a verbose msg that the machine serial is not found and will generate a random node id anyway.

For the second comment, I changed the code and updated the PR

PR #1357, opened 2019-12-09
Issue 1326 - part 2 hzn command for check both policy and userinput c…

linggao:
…ompatibility.

-  hzn and secure api for 'all' compatibility check
- retry count to retry the tranport errors for InvokeChange calls in exchange.


cli/deploycheck/allcomp.go
```diff
@@ -66,6 +264,7 @@ func newHTTPClientFactory() *config.HTTPClientFactory {
 	}
 	return &config.HTTPClientFactory{
 		NewHTTPClient: clientFunc,
+		RetryCount:    3,
```
dabooz: For CLI usage of this retry feature, I think we should default to 5 retries but make it configurable via HZN_HTTP_RETRIES env var.

Issue #1305, opened 2019-11-04
Batch device install


dabooz:
Design document is here: https://ibm.ent.box.com/notes/540559786985

- [x] The addition of artifactory to the infrastructure is out of scope of this issue.
- [x] Add new builtin property for device serial number, it is called openhorizon.hardwareId (that's a capital i). If it is possible to obtain the device serial number, we will use that value. If not, generate a crypto secure id like https://github.com/open-horizon/anax/blob/master/api/path_node.go#L128.
- [x] Unzip the tar file containing all the config files. The name of the file will be well known. The tar file will be placed on the device before this script gets control.
- [x] OCP certificate from the tar file. Configure /etc/default/horizon to point to it and make sure it is setup for any CLI commands.
- [x] The agent debs will be in the tar file, and therefore present on the device AFTER the tar is exploded.
- [x] Use mapping file (from tar) to find node name/ID to use for hzn register. Allow anax to generate a node token.
- [x] hzn register with pattern or policy, which will have been set in the node resource, so we will have to query the node resource for this info, or ensure that anax picks it up from the node during registration.
- [x] The install script can be run on a node that is already installed/configured and nothing bad happens.

PR #1358, opened 2019-12-09
Issue #1305 - add node built-in property 'openhorizon.hardwareId'

MaxMcAdam:
The added parameter on CreateNodeBuiltInPolicy is a boolean to indicate if a generated hardwareId should be included. This is only true when creating a new default policy or when updating a policy that does not already have a hardwareId defined.

### DIFF
--- a/abstractprotocol/protocol.go
+++ b/abstractprotocol/protocol.go
@@ -575,7 +575,7 @@ func addNodeBuiltInProps(pol *policy.Policy) (*policy.Policy, error) {
 	// get built-in node properties and replace the ones in the policy,
 	// the memory will be the available memory instead of total memory -- not yet,
 	//still use total memory, change to true if you want available memory
-	builtinNodePol := externalpolicy.CreateNodeBuiltInPolicy(false)
+	builtinNodePol := externalpolicy.CreateNodeBuiltInPolicy(false, false)
 	for _, prop := range builtinNodePol.Properties {
 		if err := pol.Add_Property(&prop, true); err != nil {
 			return nil, err
--- a/events/events.go
+++ b/events/events.go
@@ -810,7 +810,7 @@ func NewGovernanceWorkloadCancelationMessage(id EventId, cause EndContractCause,
 
 	return &GovernanceWorkloadCancelationMessage{
 		GovernanceMaintenanceMessage: *govMaint,
-		Cause:                        cause,
+		Cause: cause,
 	}
 }
 
--- a/exchangesync/node_policy.go
+++ b/exchangesync/node_policy.go
@@ -87,7 +87,9 @@ func GetProcessedExchangeNodePolicy(pDevice *persistence.ExchangeDevice, getExch
 		return nil, fmt.Errorf("Unable to retrieve the node policy from the exchange. Error: %v", err)
 	}
 
-	builtinPolicy := externalpolicy.CreateNodeBuiltInPolicy(false)
+	hasHwId := exchangeNodePolicy.Properties.HasProperty(externalpolicy.PROP_NODE_HARDWAREID)
+
+	builtinPolicy := externalpolicy.CreateNodeBuiltInPolicy(false, !hasHwId)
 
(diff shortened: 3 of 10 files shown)


externalpolicy/builtin_properties.go
```diff
@@ -43,6 +44,25 @@ func CreateNodeBuiltInPolicy(availableMem bool) *ExternalPolicy {
 		avail_mem = 0
 	}
 
+	hwId, err := cutil.GetMachineSerial("")
+	if hwId == "" && includeGenHwId {
+		if err != nil {
+			glog.V(2).Infof("Failed to read device serial number: %v. Proceeding with generated Id.")
```
dabooz: This fails make check, no value provided for %v.

externalpolicy/builtin_properties.go
```diff
@@ -27,7 +28,7 @@ const MAX_MEMEORY = 1048576 // the unit is MB. This is 1000G
 
 // get the node's built-in ptoperties to be used in the node policy
 // availableMem -- the total memory vs. the available memory size
-func CreateNodeBuiltInPolicy(availableMem bool) *ExternalPolicy {
+func CreateNodeBuiltInPolicy(availableMem bool, includeGenHwId bool) *ExternalPolicy {
```
dabooz: If the user does 'hzn policy update -f <new.policy.json>' where they omit the openhorizon.hardwareId in new.policy.json, and the hardwareId has to be generated (because there is no machine serial) then a new value is generated for hardwareId. That should not happen, the old value should remain the same after it is set for the first time.

externalpolicy/builtin_properties.go
```diff
@@ -43,6 +44,25 @@ func CreateNodeBuiltInPolicy(availableMem bool) *ExternalPolicy {
...
+		} else {
+			glog.V(2).Infof("Device serial number not found. Proceeding with generated Id.")
+		}
+
+		var err error
+		if hwId, err = cutil.GenerateRandomNodeId(); err != nil {
+			glog.V(1).Infof("Failed to generate device Id: %v", err)
+		}
+	} else if hwId == "" {
+		glog.V(2).Infof("Failed to read device serial number. Excluding hardwareId from build-in properties list.")
```
dabooz: typo: from built-in properties....

exchangesync/node_policy.go
```diff
@@ -330,8 +332,10 @@ func UpdateNodePolicy(pDevice *persistence.ExchangeDevice, db *bolt.DB, nodePoli
 		return fmt.Errorf("Node policy does not validate. %v", err)
 	}
 
+	hasHwId := nodePolicy.Properties.HasProperty(externalpolicy.PROP_NODE_HARDWAREID)
```
dabooz: This is the code that I think is not checking if the node already has a hardware id before generating a new one.

events/events.go
```diff
@@ -810,7 +810,7 @@ func NewGovernanceWorkloadCancelationMessage(id EventId, cause EndContractCause,
 
 	return &GovernanceWorkloadCancelationMessage{
 		GovernanceMaintenanceMessage: *govMaint,
-		Cause:                        cause,
+		Cause: cause,
```
dabooz: I think your formatter is off....you and Ling are having deuling formatters, or maybe Ling is off...

abstractprotocol/protocol.go
```diff
@@ -575,7 +575,7 @@ func addNodeBuiltInProps(pol *policy.Policy) (*policy.Policy, error) {
 	// get built-in node properties and replace the ones in the policy,
 	// the memory will be the available memory instead of total memory -- not yet,
 	//still use total memory, change to true if you want available memory
-	builtinNodePol := externalpolicy.CreateNodeBuiltInPolicy(false)
+	builtinNodePol := externalpolicy.CreateNodeBuiltInPolicy(false, false)
```
dabooz: I think this is a problem, if Create... generates a new hardwareId every time, then the hardwareId couldnt be used in a business policy constraint. I think we need the hardwareId value from the node itself at this point.

Issue #1331, opened 2019-11-19
agent install script incorrectly checks for port 80 in use


dabooz:
If there is something running on port 8080, the install script incorrectly detects this as port 80 and fails the install.

The person who reported this problem says this is the fix:
ANAX_PORT=80
echo now “Checking if the ${ANAX_PORT} is free...”
if [ ! -z “$(netstat -nlp | grep \“:$ANAX_PORT \“)” ]; then
       echo now “Something is running on ${ANAX_PORT}...”

PR #1364, opened 2019-12-10
Issue #1331 - don't match substrings when checking busy ports in agent-install

MaxMcAdam:
…stall

### DIFF
--- a/agent-install/agent-install.sh
+++ b/agent-install/agent-install.sh
@@ -34,7 +34,7 @@ function help() {
      cat << EndOfMessage
 $(basename "$0") <options> -- installing Horizon software
 where:
-    \$HZN_EXCHANGE_URL, \$HZN_FSS_CSSURL, \$HZN_ORG_ID, \$HZN_EXCHANGE_USER_AUTH variables must be defined either in a config file or environment, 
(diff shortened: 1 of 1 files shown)


agent-install/agent-install.sh
```diff
@@ -456,20 +456,20 @@ function install_linux(){
     fi
 
 	log_info "Checking if the agent port ${ANAX_PORT} is free..."
-	if [ ! -z "$(netstat -nlp | grep :$ANAX_PORT)" ]; then
+	if [ ! -z "$(netstat -nlp | grep \":$ANAX_PORT\")" ]; then
```
dabooz: I dont understand how this fix works:
$ export ABC=53
$ netstat -nlp | grep :${ABC}
(Not all processes could be identified, non-owned process info
 will not be shown, you would have to be root to see it all.)
tcp        0      0 127.0.0.53:53           0.0.0.0:*               LISTEN      -
udp        0      0 0.0.0.0:5353            0.0.0.0:*                           -
udp        0      0 127.0.0.53:53           0.0.0.0:*                           -
udp6       0      0 :::5353                 :::*                                -
$ netstat -nlp | grep \":${ABC}\"
(Not all processes could be identified, non-owned process info
 will not be shown, you would have to be root to see it all.)

MaxMcAdam: There is supposed to be a space after $ANAX_PORT. It's there now. I don't understand why that happened either

PR #1369, opened 2019-12-11
Added configurable HTTP retry count and retry interval for hzn

linggao:


agreementbot/secure_api.go
```diff
@@ -460,7 +461,7 @@ func (a *SecureAPI) authenticateWithExchange(user string, userPasswd string, msg
 	}
 
 	// Invoke the exchange API to verify the user.
-	retryCount := 2
+	retryCount := 5
```
dabooz: Use the http client factory setup earlier to get these values (retry count and interval) instead of hard coding them.

PR #1370, opened 2019-12-12
Issue #1305 - add node built-in property 'openhorizon.hardwareId'

MaxMcAdam:


producer/producer_protocol_handler.go
```diff
@@ -251,7 +251,11 @@ func (w *BaseProducerProtocolHandler) HandleProposal(ph abstractprotocol.Protoco
 			err_log_event = fmt.Sprintf("Error creating message target: %v", err)
 		} else {
 			handled = true
-			if r, err := ph.DecideOnProposal(proposal, w.ec.GetExchangeId(), exchange.GetOrg(w.ec.GetExchangeId()), runningBCs, messageTarget, w.sendMessage); err != nil {
+			producerPol, err := persistence.FindNodePolicy(w.db)
+			if err != nil {
+				glog.Errorf("Error getting node policy from db: %v", producerPol)
```
dabooz: Please follow the pattern of other glog.Errorf calls in this file

persistence/attributes.go
```diff
@@ -336,7 +336,7 @@ func AttributesToEnvvarMap(attributes []Attribute, envvars map[string]string, pr
 	writePrefix("ARCH", cutil.ArchString())
 
 	// Override with the built-in properties
-	externalPol := externalpolicy.CreateNodeBuiltInPolicy(false)
+	externalPol := externalpolicy.CreateNodeBuiltInPolicy(false, true, nil)
```
dabooz: HARDWAREID is missing from the containers, I think you need to get the node policy from the DB and pass it in and I'm seeing several hardware ids in the log again.

producer/producer_protocol_handler.go
MaxMcAdam: Fixed

persistence/attributes.go
MaxMcAdam: Fixed. HARDWAREID is now in the containers

MaxMcAdam:
There are 2 openhorizon.hardwareId values in the e2edev logs, one when anax is started, then another after the db is deleted. 
'hzn dev service start' queries the agent for the current policy. If the agent is not running, HZN_HARDWARE will not be present in the service container.

persistence/attributes.go
MaxMcAdam: Fixed

PR #1372, opened 2019-12-12
Issue 1326 - part 3 Add pattern support for user input compatibility …

linggao:
…check
1. Added support for pattern.
2. Let `hzn exchange node update` call compcheck for userInput verification for pattern case.
3. Moved share structures/functions into common package.


cli/cliutils/cliutils.go
```diff
@@ -1475,7 +1452,7 @@ func GetHTTPClient(timeout int) *http.Client {
 	TLSHandshake := dialTimeout
 	expectContinue := int(float64(requestTimeout) * 0.5)
 
-	Verbose(i18n.GetMessagePrinter().Sprintf("HTTP request timeout set to %v seconds", requestTimeout))
+	//Verbose(i18n.GetMessagePrinter().Sprintf("HTTP request timeout set to %v seconds", requestTimeout))
```
dabooz: I dont think this should be commented out. It is important debug info.

PR #1376, opened 2019-12-17
Issue #1305 - support batch node install

MaxMcAdam:
Will not overwrite configured node unless -f specified
Will not install older or equal version unless -f specified

Expecting zip file **agent-install-files.tar.gz** containing: 
**agent-install.cfg**
**agent-install.crt** 
**node-id-mapping.csv**
**\<OS\>/\<DISTRO\>/\<CODENAME\>/\<ARCH\>/.debs**

### DIFF
--- a/agent-install/agent-install.sh
+++ b/agent-install/agent-install.sh
@@ -18,13 +18,17 @@ SKIP_REGISTRATION=false
 CFG="agent-install.cfg"
 OVERWRITE=false
 HZN_NODE_POLICY=""
+AGENT_INSTALL_ZIP="agent-install-files.tar.gz"
+NODE_ID_MAPPING_FILE="node-id-mapping.csv"
+CERTIFICATE_DEFAULT="agent-install.crt"
 
 VERBOSITY=3 # Default logging verbosity
+BATCH_INSTALL=false
 
 # required parameters and their defaults
-REQUIRED_PARAMS=( "CERTIFICATE" "HZN_EXCHANGE_URL" "HZN_FSS_CSSURL" "HZN_ORG_ID" "HZN_EXCHANGE_USER_AUTH" )
+REQUIRED_PARAMS=( "HZN_EXCHANGE_URL" "HZN_FSS_CSSURL" "HZN_ORG_ID" "HZN_EXCHANGE_USER_AUTH" )
 REQUIRED_VALUE_FLAG="REQUIRED_FROM_USER"
-DEFAULTS=( "agent-install.crt" "${REQUIRED_VALUE_FLAG}" "${REQUIRED_VALUE_FLAG}" "${REQUIRED_VALUE_FLAG}" "${REQUIRED_VALUE_FLAG}" )
+DEFAULTS=( "${REQUIRED_VALUE_FLAG}" "${REQUIRED_VALUE_FLAG}" "${REQUIRED_VALUE_FLAG}" "${REQUIRED_VALUE_FLAG}" )
 
 # certificate for the CLI package on MacOS
 MAC_PACKAGE_CERT="horizon-cli.crt"
@@ -44,6 +48,8 @@ where:
     -s          - skip registration
     -v          - show version
     -l          - logging verbosity level (0-5, 5 is verbose)
+    -u          - exchange user authorization credentials
+    -f          - install older version without prompt. overwrite configured node without prompt.
 
 Example: ./$(basename "$0") -i <path_to_package(s)>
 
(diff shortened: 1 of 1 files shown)


agent-install/agent-install.sh
```diff
@@ -534,6 +539,9 @@ function install_linux(){
 				if version_gt "$AGENT_VERSION" "$PACKAGE_VERSION" ; then
 					log_notify "Installed agent ${AGENT_VERSION} is newer than the packages ${PACKAGE_VERSION}"
 					if [ ! "$OVERWRITE" = true ] ; then
+            if [ $BATCH_INSTALL ]; then
+              exit 1
```
dabooz: There are a few more places in the code that prompt for interactive response which aren't protected by the BATCH_INSTALL check.

agent-install/agent-install.sh
```diff
@@ -18,8 +18,11 @@ SKIP_REGISTRATION=false
 CFG="agent-install.cfg"
 OVERWRITE=false
 HZN_NODE_POLICY=""
+AGENT_INSTALL_ZIP="agent-install-files.tar.gz"
```
dabooz: I dont see any code to pick up the pattern or node policy from the exchange. This is the mainline path for the batch install feature. Did I miss it?

agent-install/agent-install.sh
```diff
@@ -534,6 +539,9 @@ function install_linux(){
 				if version_gt "$AGENT_VERSION" "$PACKAGE_VERSION" ; then
 					log_notify "Installed agent ${AGENT_VERSION} is newer than the packages ${PACKAGE_VERSION}"
 					if [ ! "$OVERWRITE" = true ] ; then
+            if [ $BATCH_INSTALL ]; then
```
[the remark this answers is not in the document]
MaxMcAdam: Fixed the problems that I found. The existing code uses a mix of tabs and spaces, my editor had tab = 2 space default whereas the existing code seems to have tab = 8 spaces.

agent-install/agent-install.sh
```diff
@@ -18,8 +18,11 @@ SKIP_REGISTRATION=false
...
 OVERWRITE=false
 HZN_NODE_POLICY=""
+AGENT_INSTALL_ZIP="agent-install-files.tar.gz"
+NODE_ID_MAPPING_FILE="node-id-mapping.csv"
 
 VERBOSITY=3 # Default logging verbosity
+BATCH_INSTALL=false
 
 # required parameters and their defaults
 REQUIRED_PARAMS=( "CERTIFICATE" "HZN_EXCHANGE_URL" "HZN_FSS_CSSURL" "HZN_ORG_ID" "HZN_EXCHANGE_USER_AUTH" )
```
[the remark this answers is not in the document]
MaxMcAdam: I removed the part that adds the cert to the device's trust store. Made the certificate optional, when it is there it will get set in /etc/default/horizon.

agent-install/agent-install.sh
MaxMcAdam: Fixed

agent-install/agent-install.sh
MaxMcAdam: It is there now. The script runs 'hzn exchange node create' even when the node already exists in the exchange. Is this the expected behavior? As far as I can tell, that call doesn't change node policy or pattern info in the exchange.

agent-install/agent-install.sh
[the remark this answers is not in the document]
dabooz: I am still seeing formatting problems, like line 238

agent-install/agent-install.sh
dabooz: Looks like line 790 is still not protected by a batch install check.

agent-install/agent-install.sh
dabooz: No. The problem is that the node will be precreated with a pattern (or node policy) AND the pattern or node policy will not be passed into the batch install script. The script is expected to retrieve that info from the exchange. I was looking at the code that assigns HZN_EXCHANGE_PATTERN and it seems to only assign it from input args.

agent-install/agent-install.sh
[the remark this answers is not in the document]
MaxMcAdam: Fixed that one. I don't see any others in my editor or github.

agent-install/agent-install.sh
MaxMcAdam: Fixed

agent-install/agent-install.sh
MaxMcAdam: It was doing it in the registration method before, but I moved it up to the correct spot where HZN_EXCHANGE_PATTERN is set from input args.

PR #1383, opened 2019-12-23
Issue 1292 - Add private, public key verification

jiportilla:


cli/exchange/pattern.go
```diff
@@ -224,10 +224,10 @@ func PatternPublish(org, userPw, jsonFilePath, keyFilePath, pubKeyFilePath, patN
 		cliutils.Fatal(cliutils.CLI_INPUT_ERROR, msgPrinter.Sprintf("the pattern definition (%s) must contain services, unable to proceed", patFile.Services))
 	}
 
+	keyVerified := false
```
dabooz: Why isnt keyVerified set to true right after line 270 in the original code:
keyFilePath, pubKeyFilePath = cliutils.GetSigningKeys(keyFilePath, pubKeyFilePath)

cli/cliutils/cliutils.go
```diff
@@ -1332,6 +1363,11 @@ func GetSigningKeys(privKeyFilePath, pubKeyFilePath string) (string, string) {
 	privKeyFilePath_tmp := WithDefaultEnvVar(&privKeyFilePath, "HZN_PRIVATE_KEY_FILE")
 	privKeyFilePath = VerifySigningKeyInput(*privKeyFilePath_tmp, false)
 
+	//Verify private key file format
+	if privKeyFilePath != "" {
+		verifyPrivateKeyFormat(privKeyFilePath)
+	}
+
```
dabooz: Why didnt you do the verifyPublicKeyFormat call right here?

cli/exchange/pattern.go
dabooz: I dont see changes for the pattern verify function, I think we need similar key checks.

PR #1387, opened 2020-01-02
Issue 1384 hzn mms object list -d shows mismatch destinations for mms…

LiilyZhang:
… object

This change fixed the mismatch problem, `policy-multires.tgz` and `policy-basicres.tgz` always have their destinations (shows in following gif)
Screen-Recording-2020-01-02-at-15 33 45

### DIFF
--- a/cli/sync_service/object.go
+++ b/cli/sync_service/object.go
@@ -130,15 +130,19 @@ func ObjectList(org string, userPw string, objType string, objId string, destPol
 
 		mmsObjects := make([]MMSObjectInfo, 0)
 		c := make(chan MMSObjectInfo)
-		input_chan := make(chan common.MetaData)
 
 		for i := 0; i < len(batches); i++ {
-			for _, _ = range batches[i] {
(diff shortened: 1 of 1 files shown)


cli/sync_service/object.go
```diff
@@ -130,15 +130,19 @@ func ObjectList(org string, userPw string, objType string, objId string, destPol
 
 		mmsObjects := make([]MMSObjectInfo, 0)
 		c := make(chan MMSObjectInfo)
-		input_chan := make(chan common.MetaData)
 
 		for i := 0; i < len(batches); i++ {
-			for _, _ = range batches[i] {
+			for _, obj := range batches[i] {
+				obj := obj
```
dabooz: What is this doing? Is it necessary?

dabooz: It is making a copy (it's called a shadow variable) so that the copy can be used in the go routine. For loop variables are reused on each iteration, so that would cause a problem if the go routine did not use it's own copy.

PR #1392, opened 2020-01-07
Issue 1391 - support new akka exchange

dabooz:


test/Makefile
```diff
@@ -2,7 +2,7 @@
 SHELL := /bin/bash
 
 #Versioning
-DOCKER_EXCH_TAG := 1.121.0
+DOCKER_EXCH_TAG := 2.0.7
```
dabooz: I my next PR I will be changing the code to require something newer than 2.0.7 because it will have fixes that we need.

PR #1440, opened 2020-01-22
Issue 1421-2 show env userId:pw

jiportilla:


cli/node/node.go
```diff
@@ -115,17 +114,11 @@ func Architecture() {
 
 func Env(org, userPw, exchUrl, cssUrl string) {
 	// Show hzn Environment Variables
-	mask := "******"
 	msgPrinter := i18n.GetMessagePrinter()
 	msgPrinter.Printf("Horizon Agent HZN Environment Variables are:")
 	msgPrinter.Println()
 	msgPrinter.Printf("HZN_ORG_ID: %s", org)
 	msgPrinter.Println()
-	if strings.Contains(userPw, "iamapikey:") {
```
dabooz: We need to mask out the password, just not the user.

PR #1442, opened 2020-01-22
Issue 1436 - Error getting service with multiple architectures from t…

linggao:
…he exchange for 'hzn deploycheck'

### DIFF
--- a/exchange/service.go
+++ b/exchange/service.go
@@ -461,20 +461,14 @@ func processGetSelectedServicesResponse(mURL string, mOrg string, mVersion strin
 
 	ret := map[string]ServiceDefinition{}
 
-	// If the caller wanted a specific version, check for 1 result.
 	if searchVersion != "" {
-		if len(msMetadata) != 1 {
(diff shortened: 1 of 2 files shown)


exchange/service.go
```diff
@@ -463,18 +463,11 @@ func processGetSelectedServicesResponse(mURL string, mOrg string, mVersion strin
...
-			glog.Errorf(rpclogString(fmt.Sprintf("expecting 1 service %v %v %v response: %v", mURL, mOrg, mVersion, resp)))
-			return ret, errors.New(fmt.Sprintf("expecting 1 service %v %v %v, got %v", mURL, mOrg, mVersion, len(msMetadata)))
-		} else {
-			for msId, msDef := range msMetadata {
-				glog.V(3).Infof(rpclogString(fmt.Sprintf("returning service definition %v", msDef.ShortString())))
-				ret[msId] = msDef
-				return ret, nil
-			}
-			return ret, errors.New("should not get here")
+		for msId, msDef := range msMetadata {
```
dabooz: The comment above really needs to be updated. If I am reading this correctly, a search for a specific version can return multiple services because the arch field is empty, and thus multiple services can be returned.

PR #1455, opened 2020-01-27
Fixes for 2 small cli bugs

MaxMcAdam:
bug 1) error returned by deploycheck when pattern org not specified in pattern id. Help text says it will use the user org in this case.
bug 2) querying the exchange with an empty node id when node id is retrieved from the node causing an error

### DIFF
--- a/cli/deploycheck/allcomp.go
+++ b/cli/deploycheck/allcomp.go
@@ -38,6 +38,11 @@ func AllCompatible(org string, userPw string, nodeId string, nodeArch string,
 	compCheckInput.PatternId = patternId
 	compCheckInput.Pattern = pattern
 
+	// use the user org for the patternId if it does not include an org id
+	if compCheckInput.PatternId != "" {
+		compCheckInput.PatternId = cliutils.AddOrg(userOrg, compCheckInput.PatternId)
+	}
+
 	if useNodeId {
 		// add credentials'org to node id if the node id does not have an org
 		nId = cliutils.AddOrg(userOrg, nId)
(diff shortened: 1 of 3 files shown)


cli/deploycheck/allcomp.go
```diff
@@ -261,7 +266,7 @@ func verifyCompCheckParamters(org string, userPw string, nodeId string, nodePolF
 		// Not checking the missing ones becaused it will be checked by the compcheck package.
 		checkServiceDefsForPattern(pattern, serviceDefs, svcDefFiles)
 
-		return orgToUse, *credToUse, nodeId, useNodeId, nil, pf, serviceDefs
+		return orgToUse, *credToUse, nodeIdToUse, useNodeId, nil, pf, serviceDefs
```
linggao: Please also change line 259 to use nodeIdToUse instead of nodeId. 
````
return orgToUse, *credToUse, nodeId, useNodeId, bp, nil, serviceDefs
````

File: cli/deploycheck/allcomp.go [hunk 2]
```diff
@@ -38,6 +38,11 @@ func AllCompatible(org string, userPw string, nodeId string, nodeArch string,
 	compCheckInput.PatternId = patternId
 	compCheckInput.Pattern = pattern
 
+	// use the user org for the patternId if it does not include an org id
+	if compCheckInput.PatternId != "" {
+		compCheckInput.PatternId = cliutils.AddOrg(userOrg, compCheckInput.PatternId)
+	}
+
```
linggao: Please also add the same code in userinput.go

PR #1456, opened 2020-01-27
Issue 1437 - Remote Testing Bug Fixes

njabbouribm:


test/gov/api_key.sh
```diff
@@ -22,14 +22,6 @@ ORG_ID=${AGBOT_NAME//"-agbot"/}
 
 export HZN_EXCHANGE_URL="${EXCH_APP_HOST}"
 
-# Register services via the hzn dev exchange commands
```
dabooz: Why are these removed from the api key tests? Some of what this does is very important to test with api keys.

PR #1468, opened 2020-01-29
Check type of the value field of property openhorizon.allowPrivileged when it is returned from the exchange

MaxMcAdam:


externalpolicy/builtin_properties.go
```diff
@@ -55,7 +55,19 @@ func CreateNodeBuiltInPolicy(availableMem bool, omitGenHwId bool, existingPolicy
 	if existingPolicy != nil && existingPolicy.Properties.HasProperty(PROP_NODE_PRIVILEGED) {
 		privProp, _ := existingPolicy.Properties.GetProperty(PROP_NODE_PRIVILEGED)
 		glog.Infof("Found existing policy with %v", privProp)
-		privileged = privProp.Value.(bool)
+		var ok bool
```
dabooz: Do we need the Info msg on line 57? maybe it's ok to make it a .V(5) msg.
I also noticed that the errors in here are logged as info, they should be glog.Errorf(...) like you have on line 68

MaxMcAdam: I changed it to a .V(5) . The errors here need to be .V(x) since this fuction can be called through the cli with hzn dev service start.

PR #1488, opened 2020-02-03
Issue 1478 - remove objects when unregistering

dabooz:


resource/resource_manager.go
```diff
@@ -204,7 +204,6 @@ func (r ResourceManager) StopFileSyncService() {
 			case <-stopChan:
 				glog.V(5).Infof(rmLogString(fmt.Sprintf("Embedded ESS Stop completed.")))
 				done = true
-				return
```
LiilyZhang: Question: why remove `return`?

dabooz: Because it was preventing the code below from running, which does additional cleanup of downloaded objects.

PR #1514, opened 2020-02-10
Issue 1272 - hzn register should take the same user input as the hzn …

LiilyZhang:
…userinput cmd

### DIFF
--- a/cli/register/register.go
+++ b/cli/register/register.go
@@ -13,6 +13,7 @@ import (
 	"github.com/open-horizon/anax/externalpolicy"
 	"github.com/open-horizon/anax/i18n"
 	"github.com/open-horizon/anax/persistence"
+	"github.com/open-horizon/anax/policy"
 	"github.com/open-horizon/anax/semanticversion"
 	"io/ioutil"
 	"net/http"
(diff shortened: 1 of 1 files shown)


cli/register/register.go
```diff
@@ -275,49 +288,88 @@ func DoIt(org, pattern, nodeIdTok, userPw, email, inputFile string, nodeOrgFromF
...
+				service.VersionRange = &ui.ServiceVersionRange
+				for _, input := range ui.Inputs {
+					if input.Name != "" {
+						uiMap[input.Name] = input.Value
+					}
+				}
+				attr.Mappings = &uiMap
+				attrSlice := []api.Attribute{*attr}
+				service.Attributes = &attrSlice
+				httpCode, respBody := cliutils.HorizonPutPost(http.MethodPost, "service/config", []int{201, 200, 400}, service)
```
linggao: For this case, you can just call 'node/userinput' API and pass the whole thing to it. 
Please see https://github.com/open-horizon/anax/blob/master/cli/userinput/userinput.go#L64

LiilyZhang: 👍

PR #1552, opened 2020-02-21
Issue 1298 & 860 - If 'hzn register' fails, it should automatically unregister

MaxMcAdam:


cli/register/register.go
```diff
@@ -279,11 +282,17 @@ func DoIt(org, pattern, nodeIdTok, userPw, email, inputFile string, nodeOrgFromF
...
-		//		And i can't query ALL the existing settings to make sure they are what we were going to set, because i can't query the node token.
-		cliutils.Fatal(cliutils.HTTP_ERROR, msgPrinter.Sprintf("this Horizon node is already registered or in the process of being registered. If you want to register it differently, run 'hzn unregister' first."))
+	// httpCode, _ = cliutils.HorizonPutPost(http.MethodPost, "node", []int{201, 200, cliutils.ANAX_ALREADY_CONFIGURED}, nd)
+	// if httpCode == cliutils.ANAX_ALREADY_CONFIGURED {
+	// 	// Note: I wanted to make `hzn register` idempotent, but the anax api doesn't support changing existing settings once in configuring state (to maintain internal consistency).
+	// 	//		And i can't query ALL the existing settings to make sure they are what we were going to set, because i can't query the node token.
+	// 	cliutils.Fatal(cliutils.HTTP_ERROR, msgPrinter.Sprintf("this Horizon node is already registered or in the process of being registered. If you want to register it differently, run 'hzn unregister' first."))
+	// }
+	err := CreateNode(nd, 60)
```
linggao: Could you make the waiting time configurable in the cli/cliconfig/config.go? Default can be 60.  In large systems, each call to the exchange could take very long time.  You can just use one waiting time for all the functions below.

MaxMcAdam: Added

cli/register/register.go
```diff
@@ -386,6 +406,207 @@ func DoIt(org, pattern, nodeIdTok, userPw, email, inputFile string, nodeOrgFromF
...
+		err := unregister.CheckNodeConfigState(15)
+		if err == nil {
+			break
+		}
+		if i < retries-1 {
+			msgPrinter.Printf("Error unregistering node. Retrying.")
+			msgPrinter.Println()
+			time.Sleep(time.Second * 5)
+		} else {
+			cliutils.Fatal(cliutils.CLI_GENERAL_ERROR, msgPrinter.Sprintf("Failed to unregister node."))
```
dabooz: If we get to this point, 3 retries didn't work, I think we should tell user that we are goin to try a deep clean and then call unregister.DeleteHorizonNode(false, true, 15)

cli/register/register.go
```diff
@@ -279,11 +294,17 @@ func DoIt(org, pattern, nodeIdTok, userPw, email, inputFile string, nodeOrgFromF
...
-	httpCode, _ = cliutils.HorizonPutPost(http.MethodPost, "node", []int{201, 200, cliutils.ANAX_ALREADY_CONFIGURED}, nd)
-	if httpCode == cliutils.ANAX_ALREADY_CONFIGURED {
-		// Note: I wanted to make `hzn register` idempotent, but the anax api doesn't support changing existing settings once in configuring state (to maintain internal consistency).
-		//		And i can't query ALL the existing settings to make sure they are what we were going to set, because i can't query the node token.
-		cliutils.Fatal(cliutils.HTTP_ERROR, msgPrinter.Sprintf("this Horizon node is already registered or in the process of being registered. If you want to register it differently, run 'hzn unregister' first."))
+	// httpCode, _ = cliutils.HorizonPutPost(http.MethodPost, "node", []int{201, 200, cliutils.ANAX_ALREADY_CONFIGURED}, nd)
+	// if httpCode == cliutils.ANAX_ALREADY_CONFIGURED {
+	// 	// Note: I wanted to make `hzn register` idempotent, but the anax api doesn't support changing existing settings once in configuring state (to maintain internal consistency).
```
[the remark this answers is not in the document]
MaxMcAdam: Removed

cli/register/register.go
MaxMcAdam: Added

PR #1555, opened 2020-02-24
Issue 1497 - Change the CLI to support deployment policy as an alias for business policy

MaxMcAdam:


cli/hzn.go
```diff
@@ -186,14 +186,14 @@ Environment Variables:
...
+	exAgbotListPolicyCmd := exAgbotCmd.Command("listdeploymentpol", msgPrinter.Sprintf("Display the deployment policies that this agbot is serving."))
+	exAgbotPol := exAgbotListPolicyCmd.Arg("agbot", msgPrinter.Sprintf("The agbot to list serving deployment policies for.")).Required().String()
+	exAgbotAddPolCmd := exAgbotCmd.Command("adddeploymentpol", msgPrinter.Sprintf("Add this deployment policy to the list of policies this agbot is serving. Currently only support adding all the deployment policies from an organization."))
+	exAgbotAPolAg := exAgbotAddPolCmd.Arg("agbot", msgPrinter.Sprintf("The agbot to add the deployment policy to.")).Required().String()
+	exAgbotAPPolOrg := exAgbotAddPolCmd.Arg("policyorg", msgPrinter.Sprintf("The organization of the deployment policy to add.")).Required().String()
+	exAgbotDelPolCmd := exAgbotCmd.Command("removedeploymentpol", msgPrinter.Sprintf("Remove this deployment policy from the list of policies this agbot is serving. Currently only support removing all the deployment policies from an organization."))
```
linggao: Could you also add alias for listdeploymentpol, adddeploymentpol and removedeploymentpol to support backward compatibility?

cli/hzn.go
```diff
@@ -610,13 +616,13 @@ Environment Variables:
 			credToUse = cliutils.GetExchangeAuth(*exUserPw, *exServiceAddPolicyIdTok)
 		case "service removepolicy":
 			credToUse = cliutils.GetExchangeAuth(*exUserPw, *exServiceRemovePolicyIdTok)
-		case "business listpolicy":
```
linggao: If user run `hzn exchange business listpolicy` does it go the the `deployment` case?

cli/hzn.go
MaxMcAdam: Added

cli/hzn.go
MaxMcAdam: Yes, it does

PR #1598, opened 2020-03-11
Issue 1581 - Add nodeType on node and new service attributes for services

linggao:


api/path_node.go
```diff
@@ -230,6 +235,17 @@ func CreateHorizonDevice(device *HorizonDevice,
 		return errorhandler(NewSystemError(fmt.Sprintf("error adding architecture for the exchange node. %v", err))), nil, nil
 	}
 
+	// update the node type for the exchange node
+	if *device.NodeType != exchDevice.NodeType {
+		pdr = exchange.PatchDeviceRequest{}
+		pdr.NodeType = *device.NodeType
+		if err := patchDeviceHandler(deviceId, *device.Token, &pdr); err != nil {
```
dabooz: I dont think we should be changing the node type. If there is a node with type device in the exchange, but this API call is for registering for a cluster type, then I think the registration should fail. And we should have a unit test for this case too.

cli/exchange/node.go
```diff
@@ -86,7 +86,7 @@ func NodeCreate(org, nodeIdTok, node, token, userPw, email string, arch string,
 	exchUrlBase := cliutils.GetExchangeUrl()
 
 	// Assume the user exists and try to create the node, but handle the error cases
-	putNodeReq := exchange.PutDeviceRequest{Token: nodeToken, Name: nodeName, SoftwareVersions: make(map[string]string), PublicKey: []byte(""), Arch: arch} // we only need to set the token
+	putNodeReq := exchange.PutDeviceRequest{Token: nodeToken, Name: nodeName, NodeType: nodeType, SoftwareVersions: make(map[string]string), PublicKey: []byte(""), Arch: arch} // we only need to set the token
```
dabooz: Can the user use the node create command to change the type of a node? They should not be allowed to do that.

cli/exchange/node.go
```diff
@@ -173,10 +173,17 @@ func NodeUpdate(org string, credToUse string, node string, filePath string) {
 			verifyNodeUserInput(org, credToUse, node, nId, ui)
 			patch[k] = ui
 
+		} else if k == "nodeType" {
```
dabooz: This should not be allowed.

persistence/device.go
```diff
@@ -49,6 +53,7 @@ type ExchangeDevice struct {
 	Org                string      `json:"organization"`
 	Pattern            string      `json:"pattern"`
 	Name               string      `json:"name"`
+	NodeType           string      `json:"nodeType"`
```
dabooz: If an old device is upgraded with this code, this field will be empty. I think we need should add code to SyncOnInit() to set the nodeType to device. We need to do the same thing for the exchange resource.

common/exchange_input.go
```diff
@@ -104,20 +105,22 @@ type ServiceChoiceFile struct {
...
+	URL                  string                       `json:"url"`
+	Version              string                       `json:"version"`
+	Arch                 string                       `json:"arch"`
+	Sharable             string                       `json:"sharable"`
+	MatchHardware        map[string]interface{}       `json:"matchHardware,omitempty"`
+	RequiredServices     []exchange.ServiceDependency `json:"requiredServices"`
+	UserInputs           []exchange.UserInput         `json:"userInput"`
+	Deployment           interface{}                  `json:"deployment"` // interface{} because pre-signed services can be stringified json
+	DeploymentSignature  string                       `json:"deploymentSignature,omitempty"`
+	ClusterDeployment    string                       `json:"clusterDeployment"`
```
dabooz: This needs to be an interface{} not a string

exchange/rpc.go
```diff
@@ -134,6 +119,7 @@ type Device struct {
 	Token              string             `json:"token"`
 	Name               string             `json:"name"`
 	Owner              string             `json:"owner"`
+	NodeType           string             `json:"nodeType"`
```
dabooz: I think we need a GetNodeType() on this object so that it can return device when the device type is set to device and when it's not set at all. We will have to support nodes where the nodeType is not set due to migration from 4.0 to 4.1.

exchange/rpc.go
```diff
@@ -339,14 +326,15 @@ func (p PutDeviceRequest) ShortString() string {
 
 // Please patch one field at a time.
 type PatchDeviceRequest struct {
+	NodeType           string              `json:"nodeType,omitempty"`
```
dabooz: I dont think this should be allowed

common/exchange_input.go
linggao: Fixed in the new code.

cli/exchange/node.go
linggao: I do not understand. This is the node create command. why can't user specify the node type?

cli/exchange/node.go
linggao: why? user should be allowed to change the node type if the node is not registered.

exchange/rpc.go
linggao: We should leave whatever the exchange returns to us. This way, we can decide if the field is defined or not in the code to make decisions.

persistence/device.go
linggao: Will change the code in the db read part, to assign it to "device" if is empty.

cli/exchange/node.go
dabooz: They can specify it, the question is can they change the nodetype of an existing node with this command? I think they can, which I think is a problem.

cli/exchange/node.go
linggao: removed

exchange/rpc.go
linggao: The exchange already returns "device" when the nodeType is empty.

exchange/rpc.go
linggao: removed

persistence/device.go
linggao: done

cli/exchange/node.go
linggao: changed to not allow node type change if the node eixtst.

common/exchange_input.go
```diff
@@ -49,10 +53,10 @@ func (p *PatternFile) GetServices() []exchange.ServiceReference {
 			sref := exchange.ServiceReference{}
 			sref.ServiceURL = svc.ServiceURL
 			sref.ServiceOrg = svc.ServiceOrg
-			sref.ServiceArch = svc.ServiceArch
 			sref.AgreementLess = svc.AgreementLess
 			if svc.DataVerify != nil {
 				sref.DataVerify = *svc.DataVerify
+				sref.ServiceArch = svc.ServiceArch
```
dabooz: Why is this here?

common/exchange_input.go
```diff
@@ -221,3 +241,60 @@ func (sf *ServiceFile) SupportVersionRange() {
 		}
 	}
 }
+
+// Validate a service definition.
```
dabooz: I dont see code in hzn dev CLI to do this same kind of verification. It already has verification logic, so this logic should be moved there and reused by this code, or vice versa.

cli/hzn.go
```diff
@@ -133,9 +134,9 @@ Environment Variables:
...
 	exNodeCreateNodeIdTok := exNodeCreateCmd.Flag("node-id-tok", msgPrinter.Sprintf("The Horizon Exchange node ID and token to be created. The node ID must be unique within the organization.")).Short('n').PlaceHolder("ID:TOK").String()
-	exNodeCreateNodeEmail := exNodeCreateCmd.Flag("email", msgPrinter.Sprintf("Your email address. Only needs to be specified if: the user specified in the -u flag does not exist, and you specified the 'public' org. If these things are true we will create the user and include this value as the email attribute.")).Short('e').String()
 	exNodeCreateNodeArch := exNodeCreateCmd.Flag("arch", msgPrinter.Sprintf("Your node architecture. If not specified, arch will leave blank.")).Short('a').String()
 	exNodeCreateNodeName := exNodeCreateCmd.Flag("name", msgPrinter.Sprintf("The name of your node")).Short('m').String()
+	exNodeCreateNodeType := exNodeCreateCmd.Flag("node-type", msgPrinter.Sprintf("The type of your node. The valid values are: device, cluster. If omitted, the default is device. However, the node type stay unchanged if the node already exists, only the node token will be updated.")).Short('T').String()
```
dabooz: You can specify the default value of the flag in the command definition, see line 331 default timeout flag for an example.

cli/hzn.go
```diff
@@ -316,8 +317,8 @@ Environment Variables:
 	registerCmd := app.Command("register", msgPrinter.Sprintf("Register this edge node with Horizon."))
 	nodeIdTok := registerCmd.Flag("node-id-tok", msgPrinter.Sprintf("The Horizon exchange node ID and token. The node ID must be unique within the organization. If not specified, HZN_EXCHANGE_NODE_AUTH will be used as a default. If both -n and HZN_EXCHANGE_NODE_AUTH are not specified, the node ID will be created by Horizon from the machine serial number or fully qualified hostname. If the token is not specified, Horizon will create a random token. If node resource in the exchange identified by the ID and token does not yet exist, you must also specify the -u flag so it can be created.")).Short('n').PlaceHolder("ID:TOK").String()
 	nodeName := registerCmd.Flag("name", msgPrinter.Sprintf("The name of the node. If not specified, it will be the same as the node id.")).Short('m').String()
+	nodeType := registerCmd.Flag("node-type", msgPrinter.Sprintf("The type of the node. The valid values are: device, cluster. If omitted, the default is device.")).Short('T').String()
```
dabooz: Use a default here too

cli/kube_operator/kube_deployment.go
```diff
@@ -0,0 +1,165 @@
+package kube_operator
```
dabooz: Please change the name of this package to kube_deployment to be consistent with the rest of the plugins.

api/path_node.go
```diff
@@ -169,8 +175,9 @@ func CreateHorizonDevice(device *HorizonDevice,
 	// Verify the pattern org if the patter is not in the same org as the device.
 
 	// Check the node on the exchange to see if there is a pattern already defined for the node
-	if exchDevice, err := getDeviceHandler(deviceId, *device.Token); err != nil {
-		return errorhandler(NewSystemError(fmt.Sprintf("Error getting device %v from the exchange. %v", deviceId, err))), nil, nil
+	exchDevice, err1 := getDeviceHandler(deviceId, *device.Token)
```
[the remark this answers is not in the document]
linggao: The check is done in 'hzn register'. I guess we need also do it here for the api users.

cli/hzn.go
linggao: The default is not 'device'. If the flag is missing and the node exists, the default is the same as the node type of the existing node.

cli/hzn.go
linggao: same reason.

common/exchange_input.go
linggao: must be a mistake. will fix.

common/exchange_input.go
linggao: Will change it when I start working on the 'hzn  dev'. There are more verification code to be added. In this one, the top level service is not in the exchange. I will add one to verify a service when all the services are in the exchange. etc. So this may not be the final place.

cli/hzn.go
linggao: done

cli/kube_operator/kube_deployment.go
linggao: done

api/path_node.go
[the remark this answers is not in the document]
linggao: done

common/exchange_input.go
linggao: I will open a new issue for pattern verification and 'hzn dev'.

PR #1609, opened 2020-03-17
Issue 1569 - Install script for installing the agent-in-k8s container…

LiilyZhang:
… onto the edge cluster (Part1)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,9 +25,11 @@ DEFAULT_UI = api/static/index.html
 # used for creating hzn man pages
 CLI_TEMP_EXECUTABLE := cli/hzn.tmp
 
+DOCKER_IMAGE_REPO ?= openhorizon
+
 ANAX_CONTAINER_DIR := anax-in-container
 DOCKER_IMAGE_VERSION ?= 2.23.10$(BRANCH_NAME)
-DOCKER_IMAGE_BASE = openhorizon/$(arch)_anax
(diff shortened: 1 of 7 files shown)


anax-in-k8s/Dockerfile
```diff
@@ -0,0 +1,29 @@
+FROM alpine:3.6
```
dabooz: We also need a UBI based container, again reference the css dockerfiles for how to do this.

Makefile
```diff
@@ -345,6 +355,13 @@ promote-agbot:
 
 promote-mac-pkg-and-docker: promote-mac-pkg promote-docker promote-agbot
 
+anax-k8s-image:
+	@echo "Producing ANAX K8S docker image $(ANAX_K8S_IMAGE)"
+	cp $(EXECUTABLE) $(ANAX_K8S_CONTAINER_DIR)/bin
```
dabooz: Where is the make logic to remove those binaries from the container directory?

Makefile
```diff
@@ -345,6 +355,13 @@ promote-agbot:
...
+	cp $(CLI_EXECUTABLE) $(ANAX_K8S_CONTAINER_DIR)/bin
```
dabooz: Also, you need to add that /bin directory to .gitignore. See below, the binary is in the PR, which should not be the case.

PR #1633, opened 2020-03-29
Some agent-install.sh user experience improvements

t-fine:
Updates in this PR are the checked boxes in this issue:
https://app.zenhub.com/workspaces/open-horizon-5877f2c6a8ad090f31c73853/issues/open-horizon/anax/1502

- removed creation of `~/.hzn/hzn.json` for Mac around line 542 
- added usage description of `-l`
- moved agent-install.crt to `/etc/horizon` ...this had to be `/private/etc/horizon` for Mac because docker on Mac does not have access to `/etc` by default, but it does have access to `/private` which is a symbolic link to `/etc` 
- removed ` Do you want to overwrite the current node configuration?[y/N]: `
- when rerunning agent-install.sh with the same install pkgs, it won't install them, unless a force flag is specified
- /agent-install.sh -h will exit code of 0 (if the usage is shown as a result of invalid args, then the exit code should be 1)
- getting node IP address with `ipconfig getifaddr en0` on Mac
- added the -z flag to the usage msg
- changed all occurrences of set +x to { set +x; } 2>/dev/null so it doesn't get printed

### DIFF
--- a/agent-install/agent-install.sh
+++ b/agent-install/agent-install.sh
@@ -59,8 +59,6 @@ where:
 Example: ./$(basename "$0") -i <path_to_package(s)>
 
 EndOfMessage
-
-quit 1
 }
 
 function version() {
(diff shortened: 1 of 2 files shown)


agent-install/agent-install.sh
```diff
@@ -1317,14 +1271,8 @@ function check_node_state() {
...
 			log_notify "Your node is registered"
-			echo "Do you want to overwrite the current node configuration?[y/N]:"
-			read RESPONSE
-			if [ "$RESPONSE" == 'y' ]; then
-				OVERWRITE_NODE=true
-				log_notify "The configuration will be overwritten..."
-			else
-				log_notify "You might be asked for overwrite confirmations later..."
-			fi
+			VERWRITE_NODE=true
```
dabooz: This look like a typo?

agent-install/agent-install.sh
```diff
@@ -393,6 +392,30 @@ function version_gt() {
 	test "$(printf '%s\n' "$@" | sort -V | head -n 1)" != "$1";
 }
 
+# create /etc/default/horizon file for mac or linux
+function create_config() {
+    if [[ ! -z "${HZN_EXCHANGE_URL}" ]] && [[ ! -z "${HZN_FSS_CSSURL}" ]]; then
```
bmpotter: use `-n` instead of `! -z`

agent-install/agent-install.sh
bmpotter: typo, should be `OVERWRITE_NODE`

t-fine: Fixed

agent-install/agent-install.sh
t-fine: Changed all instances of this

agent-install/agent-install.sh
t-fine: Fixed

agent-install/agent-install.sh
```diff
@@ -393,6 +392,30 @@ function version_gt() {
 	test "$(printf '%s\n' "$@" | sort -V | head -n 1)" != "$1";
 }
 
+# create /etc/default/horizon file for mac or linux
+function create_config() {
```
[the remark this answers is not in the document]
t-fine: Slacked you about this

PR #1656, opened 2020-04-14
Issue 1640 - Build ubi image for anax-in-container

linggao:


anax-in-container/Dockerfile_agbot.ubi
```diff
@@ -0,0 +1,50 @@
+FROM registry.access.redhat.com/ubi8/ubi-minimal:8.0
+
+ENV ANAX_LOG_LEVEL 3
```
dabooz, line 3: Do we need all these env vars? They look like leftovers from testing that you did?

anax-in-container/Dockerfile_agbot.ubi
```diff
@@ -0,0 +1,50 @@
...
+
+# shadow-utils contains groupadd and adduser commands
+RUN microdnf install --nodocs -y shadow-utils \
+    && microdnf install --nodocs -y openssl ca-certificates \
+    && microdnf install -y wget iptables vim-minimal procps tar gettext \
+    && wget -O jq https://github.com/stedolan/jq/releases/download/jq-1.6/jq-linux64 \
+    && chmod +x ./jq \
+    && mv jq /usr/bin
+
+# install docker cli
```
dabooz, line 23: Do we need this in the agbot?

Makefile
```diff
@@ -29,7 +29,7 @@ DOCKER_IMAGE_REPO ?= openhorizon
 ANAX_K8S_OVERRIDE ?= ""
 
 ANAX_CONTAINER_DIR := anax-in-container
-DOCKER_IMAGE_VERSION ?= 2.23.10$(BRANCH_NAME)
+DOCKER_IMAGE_VERSION ?= 2.25.3$(BRANCH_NAME)
```
dabooz, line 32: Now that we no longer install the agent via apt, I think the version of the container needs to be controlled by the deb packager project, so this version should be deleted and you need to add support for the deb packager to produce the container, similar to what is done for the CSS and agent k8s containers.

anax-in-container/config/agbot.json.tmpl
```diff
@@ -0,0 +1,34 @@
...
+        "ExchangeMessageTTL": 1800,
+        "ActiveDeviceTimeoutS": 180,
+        "DefaultWorkloadPW": "",
+        "MessageKeyPath": "msgKey",
+        "APIListen": "127.0.0.1:8080",
+        "PurgeArchivedAgreementHours": 1,
+        "CheckUpdatedPolicyS": 15,
+        "CSSURL": "${CSS_URL}",
+        "SecureAPIListenHost": "0.0.0.0",
+        "SecureAPIListenPort": "8083"
```
dabooz, line 32: I think you need the hardware arch aliases in here too.

anax-in-container/Dockerfile_agbot.ubi
linggao: Yes, it is needed by script/agbot_start.sh.  Here defines the default value. It can be overridden bu the user through `docker run -e ANAX_LOG_LEVEL=5 ...`.

anax-in-container/Dockerfile_agbot.ubi
linggao: I do not think we do. But old agbot container has docker cli installed.

anax-in-container/config/agbot.json.tmpl
linggao: done

anax-in-container/Dockerfile_agbot.ubi
linggao: The other vars like DB_CONFIG, EXCHANGE_URL etc are the default value for other users to be able to bring the  container up and running. After discussing with Carl, we have decided define a minimal set of overridable env vars for other users.

anax-in-container/Dockerfile_agbot.ubi
linggao: Removed.

Makefile
linggao: Updated the Makefile with @cgiroua. Also update Makefile and .travis.yaml for horizon-deb-packager https://github.com/open-horizon/horizon-deb-packager/pull/310.

anax-in-container/Dockerfile_agbot.ubi
linggao: Ok, after more discussion with @cgiroua , I have removed all the ENVs from here.

PR #1685, opened 2020-04-24
Issue 1592 - Agent monitors the containers started in a kubernetes cl…

MaxMcAdam:
…uster

### DIFF
--- a/governance/status.go
+++ b/governance/status.go
@@ -1,7 +1,6 @@
 package governance
 
 import (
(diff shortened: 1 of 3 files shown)


MaxMcAdam:
The error from the exchange when updating status are because the connectivity field is no longer being added to device status but is required by the exchange. Issue opened to make it optional to the exchange https://github.com/open-horizon/exchange-api/issues/346

PR #1696, opened 2020-04-29
Issue 1695 & Issue 1594

linggao:


agreementbot/policy.go
```diff
@@ -208,15 +208,15 @@ func (w *BaseAgreementWorker) HandleMMSObjectPolicy(cph ConsumerProtocolHandler,
...
+	newTargetNodes := make(map[string]string)
 	for _, agreement := range agreements {
 		for _, serviceId := range agreement.ServiceId {
 
 			if foundService, err := FindCompatibleServices(serviceId, &newPolicy, workerId, w.config.ArchSynonyms); err != nil {
 				// FindCompatibleServices logs it own errors.
 				continue
 			} else if foundService {
-				newTargetNodes[agreement.DeviceId] = true
+				newTargetNodes[agreement.DeviceId] = agreement.GetDeviceType()
```
dabooz, line 219: if you moved the device type check up here, then we would avoid calling the exchange to get the node policy and avoid the policy compat check.

linggao: Will the policy check needed for both types? I thought only the call to AssignObjectToNode() in unique for 'device' type.

dabooz: Since we know we arent going to route the object to an edge cluster node, there is no point in checking policy compatibility.

PR #1710, opened 2020-05-04
Issue 1658 - Use the namespace to create the k8s artifacts if provided

MaxMcAdam:


kube_operator/client.go
```diff
@@ -130,11 +125,29 @@ func (c KubeClient) Install(tar string, envVars map[string]string, agId string)
...
+	nsObj, namespace, err := getOperatorNamespace(apiObjMap)
+	if err != nil {
+		return err
+	}
+
+	if nsObj != nil {
+		glog.V(3).Infof(kwlog(fmt.Sprintf("attempting to create namespace %v", nsObj)))
+		_, err := c.Client.CoreV1().Namespaces().Create(nsObj)
+		if err != nil {
+			glog.V(3).Infof("Failed to create namspace %s. Continuing with installation.", nsObj.ObjectMeta.Name)
```
dabooz, line 137: Change the Infof to Warningf and use the kwlog function.

MaxMcAdam: Fixed

PR #1711, opened 2020-05-04
Swagger Validation added to make test local

JShelf:
Doesn't cause failure but outputs logs that simply show the validation errors in the swagger for ess and exchange.

### DIFF
--- a/anax-in-container/horizon-container
+++ b/anax-in-container/horizon-container
@@ -139,10 +139,23 @@ start() {
 
 	# setup directories for file sync service
 	if isMacos; then
-		fssHostSharePath=/private/var/tmp/horizon/${DOCKER_NAME}
+		fssBasePath=/private/var/tmp/horizon
 	else
-		fssHostSharePath=/var/tmp/horizon/${DOCKER_NAME}
+		fssBasePath=/private/var/tmp/horizon
 	fi
+	# set this base path to 777 to allow other users create directories under it
+	if [ -d $fssBasePath ]; then
+		# this is for backward comatibility because older version created this directory with root owned
+		if [ ! -w $fssBasePath ]; then
+			echo "sudo chmod 777 $fssBasePath"
+			sudo chmod 777 $fssBasePath
+		fi
+	else
+		mkdir -m 777 -p ${fssBasePath}
+	fi
+
+	# this is user specific and container name specific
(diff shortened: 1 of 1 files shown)


test/gov/gov-combined.sh
```diff
@@ -228,6 +228,10 @@ then
   fi
 fi
 
+if [ "${EXCH_APP_HOST}" = "http://exchange-api:8080/v1" ]; then
+	./swagger_validation.sh
```
dabooz, line 232: I think you need a $? check here to stop the test if there is a failure. Otherwise, there are so many messages that come out of e2edev that no one will ever notice the error.

test/gov/swagger_validation.sh
```diff
@@ -0,0 +1,34 @@
...
+curl -Ss ${EXCHANGE_URL}/v1/api-docs/swagger.json > /tmp/swagger.json
+
+EXCHANGEOUTPUT=$(curl -Ss -d @/tmp/swagger.json -H 'Content-Type:application/json' ${VALIDATOR_URL}/validator/debug)
+
+if [ "$EXCHANGEOUTPUT" == "{}" ]
+then
+	echo -e "No errors in the exchange swagger\n"
+else
+	echo -e "The errors in the exchange swagger are as follows:\n"
+	echo $EXCHANGEOUTPUT
```
dabooz, line 17: I think you need to exit 1 here to stop the script if there is a failure. Otherwise, there are so many messages that come out of e2edev that no one will ever notice the error. Or, you could remember the error, continue execution and run the validation on the ESS and then exit 1 from the script if either validation fails.

test/gov/swagger_validation.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/bash
+
+CONTAINERID=/swagger-validator-v2
+VALIDATOR_URL=http://172.18.0.8:8080
```
dabooz, line 4: I dont think you can count on this IP address. On my dev box, I get variations of IP@s for containers on the e2edev network. I think you are going to need code like this: https://github.com/open-horizon/anax/blob/master/test/gov/run_kube.sh#L86 to get the validation container's IP address.

test/gov/gov-combined.sh
JShelf: changed the script to external script non integrating with make test or gov combine

PR #1716, opened 2020-05-05
Issue 1623 - 'hzn exchange service publish' doesn't store the public …

linggao:
…signing key by default


cli/exchange/service.go
```diff
@@ -284,7 +284,7 @@ func SignDeployment(deployment interface{}, deploymentSignature string, baseDir
 
 	case map[string]interface{}:
 		// We know we need to sign the deployment config, so make sure a real key file was provided.
-		keyFilePath, pubKeyFilePath = cliutils.GetSigningKeys(keyFilePath, pubKeyFilePath)
+		keyFilePath, *pubKeyFilePath = cliutils.GetSigningKeys(keyFilePath, *pubKeyFilePath)
```
dabooz, line 287: GetSigningKeys should be called before SignDeployment to avoid side effects like this, where an input parameter is changed and the caller needs to see the effect of the change.

linggao: GetSigningKeys cannot move up to before SignDeployment because it is only needed when the deployment string needs to be signed. When it is not need to be signed, the keys are not needed. I made some changes so that the SignDeployment returns the publicKey only when the deployment string are signed using its private key pair.
Please review.

PR #1728, opened 2020-05-07
Issue 1602 - Remove all surfaced node errors at unregister

LiilyZhang:


governance/shutdown.go
```diff
@@ -541,6 +546,28 @@ func (w *GovernanceWorker) deleteNode(httpClientFactory *config.HTTPClientFactor
 	return nil
 }
 
+// Delete the node error from the exchange and local db
+func (w *GovernanceWorker) deleteNodeError () error {
+	glog.V(3).Infof(logString(fmt.Sprintf("deleting node surface error for node %v from exchange", w.GetExchangeId())))
+	err := exchange.DeleteSurfaceErrors(w.limitedRetryEC, w.GetExchangeId())
+	if err != nil && strings.Contains(err.Error(), "status: 404") {
```
linggao, line 553: Can you move this check inside the exchange.DeleteSurfaceErrors function? here is an example: https://github.com/open-horizon/anax/blob/master/exchange/policy.go#L174

PR #1735, opened 2020-05-08
Issue 1593 - Add built-in node policy for cluster agent

MaxMcAdam:
The new function GetClusterCountInfo is in cutil as a work-around for a package import cycle

### DIFF
--- a/api/path_service_config.go
+++ b/api/path_service_config.go
@@ -466,7 +466,7 @@ func CreateService(service *Service,
 	if err != nil {
 		glog.V(2).Infof("Failed to retrieve node policy from local db: %v", err)
 	}
-	externalPol, _ := externalpolicy.CreateNodeBuiltInPolicy(false, false, existingPol)
+	externalPol, _ := externalpolicy.CreateNodeBuiltInPolicy(false, false, existingPol, pDevice.IsEdgeCluster())
 	if externalPol != nil {
 		for _, ele := range externalPol.Properties {
 			if ele.Name == externalpolicy.PROP_NODE_CPU {
--- a/exchangesync/node_policy.go
+++ b/exchangesync/node_policy.go
@@ -91,13 +91,12 @@ func GetProcessedExchangeNodePolicy(pDevice *persistence.ExchangeDevice, getExch
 	if err != nil {
 		glog.V(2).Infof("Failed to retrieve node policy from local db: %v", err)
 	}
-
 	builtinPolicyReadOnly := &externalpolicy.ExternalPolicy{}
 	builtinPolicyReadWrite := &externalpolicy.ExternalPolicy{}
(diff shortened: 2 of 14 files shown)


api/path_service_config.go
```diff
@@ -466,7 +466,14 @@ func CreateService(service *Service,
 	if err != nil {
 		glog.V(2).Infof("Failed to retrieve node policy from local db: %v", err)
 	}
-	externalPol, _ := externalpolicy.CreateNodeBuiltInPolicy(false, false, existingPol)
+	isCluster := false
+	exchDevice, err := persistence.FindExchangeDevice(db)
```
dabooz, line 470: The variable pDevice already has device object from the local DB. I think you can just use that instead of reading it again.

exchangesync/node_policy.go
```diff
@@ -92,12 +92,19 @@ func GetProcessedExchangeNodePolicy(pDevice *persistence.ExchangeDevice, getExch
 		glog.V(2).Infof("Failed to retrieve node policy from local db: %v", err)
 	}
 
+	isCluster := false
+	exchDevice, err := persistence.FindExchangeDevice(db)
```
dabooz, line 96: Same comment, pDevice is already passed in. I dont think you need to read it again from the local DB.

exchangesync/node_policy.go
```diff
@@ -168,7 +175,14 @@ func SetDefaultNodePolicy(config *config.HorizonConfig, pDevice *persistence.Exc
 	if err != nil {
 		glog.V(2).Infof("Failed to retrieve node policy from local db: %v", err)
 	}
-	builtinNodePol, builtinNodePolReadWrite := externalpolicy.CreateNodeBuiltInPolicy(false, false, existingPol)
+	isCluster := false
+	exchDevice, err := persistence.FindExchangeDevice(db)
```
dabooz, line 179: and again, I think you can eliminate the need to read the device obj and just call pDevice.IsEdgeCluster() where you need it.

exchangesync/node_policy.go
```diff
@@ -356,7 +370,14 @@ func UpdateNodePolicy(pDevice *persistence.ExchangeDevice, db *bolt.DB, nodePoli
...
```
dabooz, line 374: and this one too

api/path_service_config.go
MaxMcAdam: fixed

exchangesync/node_policy.go
MaxMcAdam: fixed

exchangesync/node_policy.go
MaxMcAdam: fixed

exchangesync/node_policy.go
MaxMcAdam: fixed

PR #1787, opened 2020-05-21
Enhancement - agent install only select running agent pod before exec…

LiilyZhang:
… hzn command

### DIFF
--- a/agent-install/agent-install.sh
+++ b/agent-install/agent-install.sh
@@ -34,6 +34,7 @@ CONFIGMAP_NAME="openhorizon-agent-config"
 PVC_NAME="openhorizon-agent-pvc"
 RESOURCE_READY=0
 GET_RESOURCE_MAX_TRY=5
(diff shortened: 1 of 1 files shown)


agent-install/agent-install.sh
```diff
@@ -1719,7 +1719,13 @@ function check_deployment_status() {
 
 function get_pod_id() {
     log_debug "get_pod_id() begin"
-    POD_ID=$(kubectl get pod -n ${AGENT_NAMESPACE} 2> /dev/null | grep "agent-" | cut -d " " -f1 2> /dev/null)
+
+    while [[ $(kubectl get pods -n ${AGENT_NAMESPACE} -l app=agent -o 'jsonpath={..status.conditions[?(@.type=="Ready")].status}') != "True" ]]; do
```
dabooz, line 1723: You need a way to ensure that this loop doesnt run forever.

Issue #1780, opened 2020-05-19
[BUG] agent-install.sh does not check for docker pre-requisite


omega2733:
Summary: We do not check for the docker pre-requisite when installing for a device type of cluster, thus can cause the script to fail when docker is called but not installed during edge cluster agent install. We should have a pre-requisite check for this as we have pre-requisite checks for kubectl, export of variables, configuration files etc necessary to run the tool so we do not run into an unexpected failure.

Update: docker re-requisite check should also added for cluster agent install

This also further is exacerbated by issue [1779](https://github.com/open-horizon/anax/issues/1779) if hit by a user.

omega2733:
Just for clarity, this works perfectly fine for agent devices (nodes). This is only an issue for device types of "cluster"

omega2733:
I believe that Troy's PR https://github.com/open-horizon/anax/pull/1793 - which fixes the function install_linux is only called by install_${OS}. This isn't called in the install_cluster() function, I'm not sure if we wish to re-use the install_linux function or another function for cluster install pre-reqs.

LiilyZhang:
Verified with latest agent-install.sh from master

PR #1793, opened 2020-05-22
added docker check to linux install

t-fine:
Fox for bug https://github.com/open-horizon/anax/issues/1780

### DIFF
--- a/agent-install/agent-install.sh
+++ b/agent-install/agent-install.sh
@@ -644,7 +644,7 @@ function install_linux(){
     else
         log_info "docker not found, installing it..."
         set -x
-        apt-get install -y docker-ce docker-ce-cli containerd.io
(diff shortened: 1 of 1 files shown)


agent-install/agent-install.sh
```diff
@@ -637,6 +638,17 @@ function install_linux(){
         log_info "jq installed"
 	fi
 
+    log_info "Checking if docker is installed..."
+    if command -v docker >/dev/null 2>&1; then
+        log_info "docker found"
+    else
+        log_info "docker not found, installing it..."
+        set -x
+        apt-get install -y docker-ce docker-ce-cli containerd.io
```
linggao, line 647: The jq install uses apt command, here it uses apt-get command. Can it be consistent?

t-fine: Done!

PR #1796, opened 2020-05-23
Fixed external deep clean for cluster type upon unregister

linggao:
1. Fixed deep clean for cluster node.
2. Fixed unregistration hung problem when the service has dependent services. (The service has both deployment and clusterDeployment. The dependent services are for device nodes only).

### DIFF
--- a/cli/unregister/unregister.go
+++ b/cli/unregister/unregister.go
@@ -4,11 +4,11 @@ import (
 	"fmt"
 	docker "github.com/fsouza/go-dockerclient"
 	"github.com/open-horizon/anax/api"
+	"github.com/open-horizon/anax/cutil"
 	"github.com/open-horizon/anax/cli/agreement"
 	"github.com/open-horizon/anax/cli/cliutils"
 	"github.com/open-horizon/anax/i18n"
 	"github.com/open-horizon/anax/persistence"
-	"k8s.io/client-go/rest"
 	"net/http"
 	"time"
 )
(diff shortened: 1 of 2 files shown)


cli/unregister/unregister.go
```diff
@@ -151,28 +154,51 @@ func DeleteHorizonNode(removeNodeUnregister bool, deepClean bool, timeout int) e
 
 // remove local db, policy files and all the service containers
 func DeepClean() error {
+
+	// detect the node type
+	nodeType := persistence.DEVICE_TYPE_DEVICE
+	if _, err := rest.InClusterConfig(); err == nil {
```
dabooz, line 160: This check has been encapsulated into a function in cutil; cutil.NewKubeConfig()

cli/unregister/unregister.go
```diff
@@ -151,28 +154,51 @@ func DeleteHorizonNode(removeNodeUnregister bool, deepClean bool, timeout int) e
...
-	msgPrinter.Println()
-	cliutils.RunCmd(nil, "bash", "-c", "rm -f /var/horizon/*.db")
-	cliutils.RunCmd(nil, "bash", "-c", "rm -Rf /etc/horizon/policy.d/*")
+	if nodeType == persistence.DEVICE_TYPE_CLUSTER {
+		msgPrinter.Printf("Deleting local horizon DB...")
+		msgPrinter.Println()
+		cliutils.RunCmd(nil, "bash", "-c", "rm -f /var/horizon/*.db")
+		cliutils.RunCmd(nil, "bash", "-c", "rm -Rf /etc/horizon/policy.d/*")
+
+		// kill anax inside the agent container, it will get restarted by the nax.service script
```
dabooz, line 176: Does anax restart even after Lily's change to the anax.service? I dont think it will.

cli/unregister/unregister.go
linggao: done

cli/unregister/unregister.go
linggao: Yes, it still brings up anax when it is killed by kill command. Lily's fix was for pod shutdown case only.

PR #1804, opened 2020-05-26
Issue 1784 - Other Docker Registries not currently supported through …

LiilyZhang:
…agent-install.sh script

### DIFF
--- a/agent-install/agent-install.sh
+++ b/agent-install/agent-install.sh
@@ -1486,10 +1486,22 @@ function getImageInfo() {
 
 function pushImageToEdgeClusterRegistry() {
     log_debug "pushImageToEdgeClusterRegistry() begin"
-    echo "$EDGE_CLUSTER_REGISTRY_TOKEN" | docker login -u $EDGE_CLUSTER_REGISTRY_USERNAME --password-stdin
-    if [ $? -ne 0 ]; then
-        log_notify "Failed to login to edge cluster's registry, exiting..."
-        exit 1
+
+    # split $IMAGE_ON_EDGE_CLUSTER_REGISTRY by "/"
+    parts=$(echo $IMAGE_ON_EDGE_CLUSTER_REGISTRY|awk -F'/' '{print NF}')
+    if [ "$parts" == "3" ]; then
+        EDGE_CLUSTER_REGISTRY_HOST=$(echo $IMAGE_ON_EDGE_CLUSTER_REGISTRY|awk -F'/' '{print $1}')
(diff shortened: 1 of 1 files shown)


agent-install/agent-install.sh
```diff
@@ -1486,10 +1486,22 @@ function getImageInfo() {
...
+    if [ "$parts" == "3" ]; then
+        EDGE_CLUSTER_REGISTRY_HOST=$(echo $IMAGE_ON_EDGE_CLUSTER_REGISTRY|awk -F'/' '{print $1}')
+        log_notify "Edge cluster registy host: $EDGE_CLUSTER_REGISTRY_HOST"
+        if [ -z $EDGE_CLUSTER_REGISTRY_USERNAME ] && [ -z $EDGE_CLUSTER_REGISTRY_TOKEN ]; then
+                docker login $EDGE_CLUSTER_REGISTRY_HOST
+        elif
+                echo "$EDGE_CLUSTER_REGISTRY_TOKEN" | docker login -u $EDGE_CLUSTER_REGISTRY_USERNAME --password-stdin $EDGE_CLUSTER_REGISTRY_HOST
+        fi
+
+        if [ $? -ne 0 ]; then
```
bmpotter, line 1501: Because `set -e` is used throughout this script, if any command fails (non-zero exit code), `agent-install.sh` will exit immediately. So this check will never be true. (Note: i don't like `set -e` and we are going to change it at the beginning of the next release.)

So you have 2 choices:
1. Surround this whole section with `set +e` at the beginning and then `set -e` at the end
1. Or remove this `if [ $? -ne 0 ]; then ... fi` statement and just let `agent-install.sh` exit if `docker login` fails (its stderr goes to the terminal, so the user will see the error msg)

You have several other checks like this in other functions that should be fixed too.

And earlier in the script, you have a line `docker --help > /dev/null 2>&1` that also needs to surrounded with `set +e` beforehand and then `set -e` afterward. Because without that `agent-install.sh` exits silently if docker isn't installed. I hit that yesterday.

LiilyZhang: @bmpotter `set +e` is set before calling the install_cluster function at https://github.com/open-horizon/anax/blob/2c758d0d2c29d1ef16e747b7af139c0c7bda91c7/agent-install/agent-install.sh#L1909
So do you want me to remove `set +e` at line 1909 and add `set +e`, `set -e` separately?

bmpotter: It is fine how it is

PR #1837, opened 2020-06-01
Hold until 4.2: Updating cluster agent using agent-install.sh

LiilyZhang:
use `./agent-install.sh -D cluster -l 5` to update. Make sure to use same namespace 
If the old agent pod is not in “RUNNING” statue, will show a message to let people uninstall agent first.


LiilyZhang:
This is invalid for current anax base code. Will make changes and apply to current code base in a new branch/PR. Closed

PR #1852, opened 2020-06-05
Add parameters to choose if force delete in agent-uninstall.sh

LiilyZhang:


agent-install/agent-uninstall.sh
```diff
@@ -267,15 +271,45 @@ function deleteAgentResources() {
 
     set +e
     log_info "Deleting agent deployment..."
-    $KUBECTL delete deployment $DEPLOYMENT_NAME -n $AGENT_NAMESPACE --force=true --grace-period=0
+    
+    if [ "$USE_DELETE_FORCE" != true ]; then
+        $KUBECTL delete deployment $DEPLOYMENT_NAME -n $AGENT_NAMESPACE --grace-period=10
```
bmpotter, line 276: Please make the timeout (10 in this case) configurable via an env var `DELETE_TIMEOUT`

agent-install/agent-uninstall.sh
```diff
@@ -85,6 +86,7 @@ function show_config() {
     echo "HZN_EXCHANGE_USER_AUTH: <specified>"
     echo "Delete node: ${DELETE_EX_NODE}"
     echo "Agent namespace to uninstall: ${AGENT_NAMESPACE}"
+    echo "Force delete cluster resources: ${USE_DELETE_FORCE}"
```
bmpotter, line 89: Initialize `USE_DELETE_FORCE` to `false`. That's more clear than empty

PR #1869, opened 2020-06-15
Issue 1422 - hzn register waits for all services in a pattern

lichristina118:
This fixes issue 1422 and 1423.

### DIFF
--- a/cli/hzn.go
+++ b/cli/hzn.go
@@ -332,8 +332,8 @@ Environment Variables:
 	nodepolicyFlag := registerCmd.Flag("policy", msgPrinter.Sprintf("A JSON file that sets or overrides the node policy for this node that will be used for policy based agreement negotiation. Mutually exclusive with -p argument.")).String()
 	org := registerCmd.Arg("nodeorg", msgPrinter.Sprintf("The Horizon exchange organization ID that the node should be registered in. Mutually exclusive with -o and -p.")).String()
 	pattern := registerCmd.Arg("pattern", msgPrinter.Sprintf("The Horizon exchange pattern that describes what workloads that should be deployed to this node. If the pattern is from a different organization than the node, use the 'other_org/pattern' format. Mutually exclusive with -o, -p and --policy.")).String()
-	waitServiceFlag := registerCmd.Flag("service", msgPrinter.Sprintf("Wait for the named service to start executing on this node.")).Short('s').String()
-	waitServiceOrgFlag := registerCmd.Flag("serviceorg", msgPrinter.Sprintf("The org of the service to wait for on this node. If omitted, the node's org is used.")).String()
+	waitServiceFlag := registerCmd.Flag("service", msgPrinter.Sprintf("Wait for the named service to start executing on this node. When registering with a pattern, use '*' to watch all the services in the pattern. When registering with a policy, '*' is not a valid value for -s.")).Short('s').String()
+	waitServiceOrgFlag := registerCmd.Flag("serviceorg", msgPrinter.Sprintf("The org of the service to wait for on this node. If '-s *' is specified, then --serviceorg must be omitted.")).String()
 	waitTimeoutFlag := registerCmd.Flag("timeout", msgPrinter.Sprintf("The number of seconds for the --service to start. The default is 60 seconds, beginning when registration is successful. Ignored if --service is not specified.")).Short('t').Default("60").Int()
 
 	keyCmd := app.Command("key", msgPrinter.Sprintf("List and manage keys for signing and verifying services."))
--- a/cli/register/register.go
+++ b/cli/register/register.go
@@ -97,7 +97,7 @@ func DoIt(org, pattern, nodeIdTok, userPw, inputFile string, nodeOrgFromFlag str
 	msgPrinter := i18n.GetMessagePrinter()
 
 	// check the input
-	org, pattern = verifyRegisterParamters(org, pattern, nodeOrgFromFlag, patternFromFlag)
+	org, pattern, waitService, waitOrg = verifyRegisterParamters(org, pattern, nodeOrgFromFlag, patternFromFlag, waitService, waitOrg)
 
 	cliutils.SetWhetherUsingApiKey(nodeIdTok) // if we have to use userPw later in NodeCreate(), it will set this appropriately for userPw
 	// Read input file 1st, so we don't get half way thru registration before finding the problem
@@ -267,6 +267,7 @@ func DoIt(org, pattern, nodeIdTok, userPw, inputFile string, nodeOrgFromFlag str
 		}
 	}
 
+	checkPattern := false
 	// Use the exchange node pattern if any
 	if pattern == "" {
 		if exchangePattern == "" {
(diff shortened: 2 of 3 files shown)


cli/register/servicewait.go
```diff
@@ -24,115 +105,193 @@ func WaitForService(org string, waitService string, waitTimeout int, pattern str
...
+			}
 		}
 	}
 
-	// 3. The service might not even be there at all.
-	if !found {
-		msgPrinter.Printf("Service %v/%v is not deployed to the node, continuing analysis.", org, waitService)
+	// Print the status of each service
+	if len(failedArr) > 0 {
+		msgPrinter.Printf("The following services execution failed:")
```
dabooz, line 257: "The following services failed during execution:"

cli/register/servicewait.go
```diff
@@ -24,115 +105,193 @@ func WaitForService(org string, waitService string, waitTimeout int, pattern str
...
+	if len(notDeployedArr) > 0 {
+		msgPrinter.Printf("The following services are not deployed to the node, continuing analysis:")
+		msgPrinter.Println()
+		for _, ss := range notDeployedArr {
+			msgPrinter.Printf("\t%v/%v", ss.org, ss.name)
 			msgPrinter.Println()
 		}
 	}
 
+	// 4. Are there any agreements being made? Check for only non-archived agreements. Skip this if we know the service failed
```
dabooz, line 283: I think this section should only run when delayedArr OR notDeployedArr are not empty. I think that's consistent with the behavior of the code before your change.

cli/register/servicewait.go
```diff
@@ -148,32 +307,44 @@ func WaitForService(org string, waitService string, waitTimeout int, pattern str
 		}
 		msgPrinter.Println()
 	} else {
-		msgPrinter.Printf("The following errors were found in the node's event log and are related to %v/%v. Use 'hzn eventlog list -s severity=error -l' to see the full detail of the errors.", org, waitService)
-		msgPrinter.Println()
+		for _, ss := range servSpecArr {
```
dabooz, line 310: I think we need to check ss.Status before looking for event log messages. If the service started (i.e. the status is good) then we dont need to check the event log for errors related to that service.

cli/register/servicewait.go
```diff
@@ -24,115 +105,193 @@ func WaitForService(org string, waitService string, waitTimeout int, pattern str
...
-				// The service could fail quickly if we happened to catch it just as it was starting, so make sure
-				// the service stays up.
-				serviceUp += 1
+		// for each service that we are monitoring, check if they are up
+		for i, ss := range servSpecArr {
+			if ss.status == 0 {
+				for _, serviceInstance := range instances {
+					if serviceInstance.SpecRef == ss.name && (serviceInstance.Org == ss.org || ss.org == "*") {
+						if ss.org == "*" {
+							if ServiceInserted(servSpecArr, ss.name, serviceInstance.Org) {
```
dabooz, line 154: Can you explain what's going on in this code, I'm not following it. You're modifying the current array elements and then appending the current element to the end....

lichristina118: This is the policy case where org is '*'. Since we do not know how many agreements there will be, we insert the ones that match into the beginning of the list and add the element with '*' as org to the end so we can keep adding new agreements that match. After the loops are done, we can simply remove the last element in the array (the '*' org).

cli/register/register.go
```diff
@@ -653,7 +660,22 @@ func verifyRegisterParamters(org, pattern, nodeOrgFromFlag string, patternFromFl
 	if org == "" {
 		cliutils.Fatal(cliutils.CLI_INPUT_ERROR, msgPrinter.Sprintf("Please specify the node organization id."))
 	}
-	return org, pattern
+
+	if waitService == "" && waitOrg != "" {
+		cliutils.Fatal(cliutils.CLI_INPUT_ERROR, msgPrinter.Sprintf("-s must be specified if --serviceorg is specified"))
+	}
+
+	// if waitOrg is omitted, default to '*'
```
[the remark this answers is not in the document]
lichristina118: I don't quite understand this comment, but in my design for the pattern case, I accept the following combinations:
- waitService = service name, waitOrg = org name
- waitService = *, waitOrg = org name
- waitService = service name, waitOrg = * or omitted
- waitService = *, waitOrg = * or omitted 
Not accepting it when waitService omitted, waitOrg not empty

cli/hzn.go
```diff
@@ -332,8 +332,8 @@ Environment Variables:
...
 	org := registerCmd.Arg("nodeorg", msgPrinter.Sprintf("The Horizon exchange organization ID that the node should be registered in. Mutually exclusive with -o and -p.")).String()
 	pattern := registerCmd.Arg("pattern", msgPrinter.Sprintf("The Horizon exchange pattern that describes what workloads that should be deployed to this node. If the pattern is from a different organization than the node, use the 'other_org/pattern' format. Mutually exclusive with -o, -p and --policy.")).String()
-	waitServiceFlag := registerCmd.Flag("service", msgPrinter.Sprintf("Wait for the named service to start executing on this node.")).Short('s').String()
-	waitServiceOrgFlag := registerCmd.Flag("serviceorg", msgPrinter.Sprintf("The org of the service to wait for on this node. If omitted, the node's org is used.")).String()
+	waitServiceFlag := registerCmd.Flag("service", msgPrinter.Sprintf("Wait for the named service to start executing on this node. Use '*' to watch all the services in the pattern.")).Short('s').String()
```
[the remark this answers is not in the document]
lichristina118: Done.

cli/hzn.go
```diff
@@ -332,8 +332,8 @@ Environment Variables:
...
+	waitServiceOrgFlag := registerCmd.Flag("serviceorg", msgPrinter.Sprintf("The org of the service to wait for on this node. If '*' is specified, watch the services defined by -s for all orgs. If omitted, default to '*'")).String()
```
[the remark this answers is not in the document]
lichristina118: Done.

cli/register/register.go
```diff
@@ -653,7 +660,22 @@ func verifyRegisterParamters(org, pattern, nodeOrgFromFlag string, patternFromFl
...
+	if waitOrg == "" {
+		waitOrg = "*"
+	}
+
+	// for the policy case, waitService = '*' is not supported
+	if pattern == "" && waitService == "*" {
+		cliutils.Fatal(cliutils.CLI_INPUT_ERROR, msgPrinter.Sprintf("For the policy case, '*' is not supported for -s."))
```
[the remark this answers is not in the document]
lichristina118: Done.

cli/register/register.go
```diff
@@ -281,25 +282,31 @@ func DoIt(org, pattern, nodeIdTok, userPw, inputFile string, nodeOrgFromFlag str
 			msgPrinter.Printf("Pattern %s defined for the node on the Exchange. Will proceeed with this pattern.", exchangePattern)
```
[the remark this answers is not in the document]
lichristina118: Done.

cli/register/servicewait.go
```diff
@@ -24,115 +105,193 @@ func WaitForService(org string, waitService string, waitTimeout int, pattern str
...
+		}
+	} else {
+		servSpecArr = append(servSpecArr, serviceSpec{name: waitService, org: org, status: 0, serviceUp: 0})
+	}
+
 	// 1. Wait for the /service API to return a service with url that matches the input
 	// 2. While waiting, report when at least 1 agreement is formed
 
-	msgPrinter.Printf("Waiting for up to %v seconds for service %v/%v to start...", waitTimeout, org, waitService)
+	msgPrinter.Printf("Waiting for up to %v seconds for following services to start...", waitTimeout)
```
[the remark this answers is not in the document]
lichristina118: Done.

cli/register/servicewait.go
lichristina118: Done.

cli/register/servicewait.go
lichristina118: Done.

cli/register/servicewait.go
lichristina118: Not sure what you mean by the service is good, but I made changes based on what I think you're trying to say. Let me know if it's not quite what you want.

cli/register/servicewait.go
```diff
@@ -24,115 +105,193 @@ func WaitForService(org string, waitService string, waitTimeout int, pattern str
 		cliutils.Fatal(cliutils.CLI_INPUT_ERROR, msgPrinter.Sprintf("--timeout must be a positive integer."))
 	}
 
+	servSpecArr := []serviceSpec{}
+	if pattern != "" {
+		for _, s := range pat.Services {
+			if (waitService == "*" || waitService == s.ServiceURL) && (org == "*" || org == s.ServiceOrg) && (s.ServiceArch == "*" || nodeArch == s.ServiceArch) {
+				if len(s.ServiceVersions) > 0 && CheckService(s.ServiceURL, s.ServiceOrg, nodeArch, s.ServiceVersions[0].Version, nodeType, userOrg, userPw) {
```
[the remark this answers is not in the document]
lichristina118: Since CheckService only makes sure that the service type and node type matches, are you saying that different service version might have different service types?

cli/register/register.go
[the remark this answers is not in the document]
dabooz: It looks like github markdown lost an important aspect of my comment. It should have started with 'if -s is "\*" and there is a pattern...
I'm trying to change your design to simplify the end user experience. For the case where there is a pattern and -s is set to wait for all services, then the user should be able to omit --serviceorg. I am saying that we should require that they omit it, i.e. it's an error to specify -s "\*" and --serviceorg when there is a pattern. This change has a subtle impact on a few places in the code.

cli/register/servicewait.go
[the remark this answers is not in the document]
dabooz: ah, my comment is invalid, you are correct. I made an incorrect assumption about why you are passing service version, and neglected to check what CheckService was actually doing. Please add a comment before line 112, something like: Collecting the set of services which apply to this node type and architecture.

cli/register/servicewait.go
dabooz: Perhaps another misunderstanding on my part. ss.Status seems to indicate that the service was started or not (ss.status == -1). We only want to extract log records from the error log and display for the cases where the service fails to get into execution. Perhaps ss.status is the wrong way to discriminate these cases?

cli/register/servicewait.go
dabooz: But in the policy case, -s "\*" is not supported, the user has to provide a service name. Org is optional in this case. So there is only 1 service that we need to monitor for.

cli/register/register.go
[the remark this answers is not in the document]
lichristina118: Done.

cli/register/servicewait.go
lichristina118: Done. Added a check for ss.status != 1.

cli/register/servicewait.go
lichristina118: Fixed it. I removed the need for the serviceInstance function and the place where we removed the last element of servSpecArr. In this case, we simply need to catch one that matches.

PR #1872, opened 2020-06-16
Issue 1860 - 'hzn service log' does not accept fully-qualified servic…

lichristina118:
…e name

### DIFF
--- a/cli/service/service.go
+++ b/cli/service/service.go
@@ -7,6 +7,7 @@ import (
 	"github.com/open-horizon/anax/api"
 	"github.com/open-horizon/anax/apicommon"
 	"github.com/open-horizon/anax/cli/cliutils"
+	"github.com/open-horizon/anax/cutil"
 	"github.com/open-horizon/anax/exchange"
 	"github.com/open-horizon/anax/i18n"
 	"github.com/open-horizon/anax/persistence"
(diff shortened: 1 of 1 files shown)


cli/service/service.go
```diff
@@ -105,10 +106,11 @@ func Log(serviceName string, tailing bool) {
 		serviceFound := false
 		var instanceId string
 		for _, serviceInstance := range runningServices.Instances["active"] {
-			if strings.Contains(serviceInstance.SpecRef, refUrl) {
+			org, name := cutil.SplitOrgSpecUrl(refUrl)
```
dabooz, line 109: The original code allows the user to specify a substring of the service name and the command still works. We need to retain that capability.

lichristina118: Done. I rewrote the condition to be more strict in the case that the org and name is inputted and only need a substring in the case that only a name is inputted.

cli/service/service.go
```diff
@@ -105,10 +106,11 @@ func Log(serviceName string, tailing bool) {
...
+			if (serviceInstance.SpecRef == name && serviceInstance.Org == org) || strings.Contains(serviceInstance.SpecRef, refUrl) {
 				instanceId = serviceInstance.InstanceId
 				serviceFound = true
-				msgPrinter.Printf("Displaying log messages for service %v with service id %v.", serviceInstance.SpecRef, instanceId)
+				msgPrinter.Printf("Displaying log messages for service %v with service id %v.", refUrl, instanceId)
```
dabooz, line 113: I think this msgs needs to specify serviceInstance.SpecRef so that the user knows which service we found. refUrl might be a substring of the full service name.

lichristina118: Done.

PR #1893, opened 2020-06-29
Issue 1811 - Change agent mount volume in e2edev to use persistent vo…

LiilyZhang:
…lume claim


test/docker/fs/etc/agent-in-kube/persistent-claim.yaml
```diff
@@ -0,0 +1,12 @@
...
+metadata:
+  name: agent-pvc-horizon
+  namespace: openhorizon-agent
+spec:
+  storageClassName: "microk8s-hostpath"
+  accessModes:
+    - ReadWriteOnce
+  resources:
+    requests:
+      storage: 2Gi
```
dabooz, line 12: Does it need to be this big?

PR #1894, opened 2020-06-30
Issue 1888 - agbot failed to cache the new patterns if the there are …

linggao:
…no patterns in a org at the start up
I tested the code with the following cases:

1. start agbot when pattern orgs and policy orgs are added added for the agbot. Add first pattern for the org. Then register a node with the pattern, agreements formed.
2. same as above except that it is for the policy case. The agreements formed.
3. start agbot when there are no pattern orgs for the agbot. Add pattern orgs for the agbot. Add first pattern for the org. Then register a node with the pattern, agreements formed.
4. same as above except that it is for the policy case. The agreements formed.
5. start agbot when pattern orgs are added for the agbot and there are patterns in the org. Delete all the patterns from the org. Then add a pattern in the org. Register a node, agreements formed.
6. same as above except that it is for the policy case. The agreements formed.


linggao:
I tested the deletion case already.

PR #1911, opened 2020-07-15
Issue 1510 - Multiple Agents

njabbouribm:


test/gov/multiple_agents.sh
```diff
@@ -0,0 +1,37 @@
...
+  NUM_AGENTS=${MUL_AGENTS}
+else
+  NUM_AGENTS=4
+fi
+while [ ${COUNTER} -lt ${NUM_AGENTS} ]; do
+  echo ${COUNTER}
+  AGENT_PORT=$((8512 + ${COUNTER}))
+  DEVICE_NUM=$((6 + ${COUNTER}))
+  echo ${AGENT_PORT}
+  echo ${DEVICE_NUM}
```
dabooz, line 13: Instead of these 3 echos filling up the console, it might be better to just echo the generated /tmp/horizon file after you've completed it.

test/gov/multiple_agents.sh
```diff
@@ -0,0 +1,37 @@
...
+  export HZN_EXCHANGE_URL=${EXCH_APP_HOST};
+  export HZN_EXCHANGE_USER_AUTH="e2edev@somecomp.com/e2edevadmin:e2edevadminpw";
+  export HZN_FSS_CSSURL=${CSS_URL};
+  export HZN_AIC="/tmp/anax-in-container/horizon-container";
+  export HORIZON_NUM=${DEVICE_NUM};
+  #export HC_DOCKER_TAG=2.26.0
+  ${HZN_AIC} start  ${HORIZON_NUM} /tmp/horizon
+  sleep 10
+  docker cp /tmp/exports_list horizon$HORIZON_NUM:/tmp/exports_list
+  docker cp /root/. horizon$HORIZON_NUM:/root/.
```
dabooz, line 33: hmmm...you are doing this copy to get the run_agent.sh script into the anax container, but the container already knows how run anax, so why do you need to do something different?

test/gov/run_agent.sh
```diff
@@ -0,0 +1,25 @@
...
+export ANAX_API="http://localhost:${HZN_AGENT_PORT}"
+export EXCH="${EXCH_APP_HOST}"
+
+export PATTERN="sall"
+export NOK8S=1
+export constraint2="NOK8S == true"
+export PARTNERID="$DEVICE_ORG/anaxdevice${DEVICE_NUM}"
+export HZN_MGMT_HUB_CERT_PATH="/tmp/css.crt"
+
+./apireg.sh > agent_debug.log
```
dabooz, line 25: I dont think this is the right way to configure the agents. The apireg way of configuring an agent uses the undocumented anax API. We do it this way with the embedded agent so that the API gets tested, but its not realistic from a customer perspective, I thikn these additional agents should be setup the way a customer would do it to improve the accuracy and completeness of the test.

test/gov/service_apireg.sh
```diff
@@ -1686,6 +1687,350 @@ read -d '' msdef <<EOF
   ]
 }
 EOF
+else
+read -d '' msdef <<EOF
```
dabooz, line 1691: Why is this duplicated? Do multi agents only work for the sall pattern test?

test/gov/verify_agreements.sh
```diff
@@ -9,8 +9,43 @@ echo -e "${PREFIX} starting"
 function verifyAgreements {
         # Wait until there are agreements
         TARGET_NUM_AG=1
+        MUL_NUM_AG_1=0
+        MUL_NUM_AG_2=0
+        MUL_NUM_AG_3=0
+        MUL_NUM_AG_4=0
         if [ "${HZN_REG_TEST}" != "1" ]; then
             if [ "${PATTERN}" == "sall" ]; then
+              case ${MUL_AGENTS} in
```
dabooz, line 18: there is way too much duplicated code here, we need to find a better way to deal with this.

PR #1934, opened 2020-07-30
Issue 1863 - Implement a new algorithm for timing calls to the change…

linggao:
…s api

### DIFF
--- a/changes/commands.go
+++ b/changes/commands.go
@@ -31,13 +31,15 @@ func NewAgreementCommand() *AgreementCommand {
 	return &AgreementCommand{}
 }
 
-type ResetIntervalCommand struct {
+type UpdateIntervalCommand struct {
+	// the update type. It can be RESET, set to MEDIAN which usually is (max+min)/2
+	UpdateType string
 }
 
-func (c ResetIntervalCommand) ShortString() string {
-	return fmt.Sprintf("ResetIntervalCommand")
+func (c UpdateIntervalCommand) ShortString() string {
+	return fmt.Sprintf("UpdateIntervalCommand: UpdateType: %v", c.UpdateType)
 }
 
-func NewResetIntervalCommand() *ResetIntervalCommand {
-	return &ResetIntervalCommand{}
+func NewUpdateIntervalCommand(updateType string) *UpdateIntervalCommand {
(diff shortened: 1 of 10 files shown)


changes/commands.go
```diff
@@ -31,13 +31,15 @@ func NewAgreementCommand() *AgreementCommand {
...
-type ResetIntervalCommand struct {
+type UpdateIntervalCommand struct {
+	// the update type. It can be RESET, set to MEDIAN which usually is (max+min)/2
+	UpdateType string
 }
 
-func (c ResetIntervalCommand) ShortString() string {
-	return fmt.Sprintf("ResetIntervalCommand")
+func (c UpdateIntervalCommand) ShortString() string {
+	return fmt.Sprintf("UpdateIntervalCommand: UpdateType: $v", c.UpdateType)
```
dabooz, line 40: That should be %v

exchange/rpc.go
```diff
@@ -390,8 +390,8 @@ func (p PutDeviceRequest) ShortString() string {
 // Please patch one field at a time.
 type PatchDeviceRequest struct {
 	UserInput          *[]policy.UserInput `json:"userInput,omitempty"`
-	Pattern            string              `json:"pattern,omitempty"`
-	Arch               string              `json:"arch,omitempty"`
+	Pattern            *string             `json:"pattern,omitempty"`
+	Arch               *string             `json:"arch,omitempty"`
```
dabooz, line 394: Dont you have to update the String() function to correctly dump the string values, since you've changed them to pointers?

linggao: These changes (changing string to string pointer) actually were for another PR which sets the pattern to an empty string at the shutdown process. The old way uses PUT, I switch it to PATCH, so we need to change it to *string to distinguish between string not set and set for an empty string. Here I simply copy the code so that I do not have to modify the String() and ShortString() functions after the merge.

changes/commands.go
linggao: oops, will fix. :-)

linggao: done

exchange/rpc.go
linggao: done

PR #1944, opened 2020-08-07
1903 - hzn mms object download does not handle existing file correctly

sirotsinskuy:


cli/sync_service/download.go
```diff
@@ -55,6 +55,8 @@ func ObjectDownLoad(org string, userPw string, objType string, objId string, fil
 					filePath = filePath + "/"
 				}
 				fileName = fmt.Sprintf("%s%s_%s", filePath, objType, objId)
+			} else {
+				fileName = filePath
```
dabooz, line 59: I think you will need to update the code starting at https://github.com/open-horizon/anax/blob/master/cli/sync_service/download.go#L62 in order to handle the case where the file already exists, including a new message and a flag to overwrite the existing file as per the issue description.

sirotsinskuy: Sure. Updated PR with adding the new flag.
Previously I was only fixing a bug (that's first part of issue - "At the very least the error msg should say the file already exists"), so after my previous change user would receive that message. I thought that if this issue has an `bug` label we'll need only fix the bug, and if we need to implement a new flat - there will be one more `feature` issue for that.

cli/sync_service/download.go
```diff
@@ -55,12 +55,16 @@ func ObjectDownLoad(org string, userPw string, objType string, objId string, fil
...
 			}
 		}
 	}
 
-	if _, err := os.Stat(fileName); err == nil {
-		cliutils.Fatal(cliutils.CLI_INPUT_ERROR, msgPrinter.Sprintf("File %s already exists. Please specify different file path or file name. Or remove file and retry object download", fileName))
+	if !overwrite {
+		if _, err := os.Stat(fileName); err == nil {
+			cliutils.Fatal(cliutils.CLI_INPUT_ERROR, msgPrinter.Sprintf("File %s already exists. Please specify different file path or file name. Or use `overwrite` flag to overwrite existing file.", fileName))
```
dabooz, line 66: small tweak to the message, please change to:
File %s already exists. Please specify a different file path or file name. To overwrite the existing file, use the '--overwrite' flag.

cli/sync_service/download.go
dabooz: I understand. In this project, we don't make a strong distinction between bug and feature, especially when it comes to user facing interfaces like the CLI.

cli/sync_service/download.go
sirotsinskuy: Sure, message has been updated

PR #1945, opened 2020-08-07
Updates to e2edev test scripts

bencourliss:
Modifications to allow e2edev test scripts to use prebuilt binaries instead of building Anax for every run. This will help in CI pipelines where we've already built the binaries, deployed to a test environment, and want to run the e2edev tests against that test environment.

### DIFF
--- a/test/Makefile
+++ b/test/Makefile
@@ -54,6 +54,12 @@ E2EDEV_HOST_IP := $(shell hostname -I | awk '{print $$1}')
 AGBOT_SAPI_URL ?= https://localhost:8083
 ICP_HOST_IP ?= 0
 
+export PREBUILT_DOCKER_REG_URL ?= ""
+export PREBUILT_DOCKER_REG_USER ?= ""
+export PREBUILT_DOCKER_REG_PW ?= ""
+export PREBUILT_ANAX_VERSION ?= nightly
+export PREBUILT_ESS_VERSION ?= nightly
+
 ifndef verbose
 .SILENT:
 endif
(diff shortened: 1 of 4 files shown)


test/Makefile
```diff
@@ -54,6 +54,17 @@ E2EDEV_HOST_IP := $(shell hostname -I | awk '{print $$1}')
...
 ICP_HOST_IP ?= 0
 
+PREBUILT_DOCKER_REG_URL ?= ""
+PREBUILT_DOCKER_REG_USER ?= ""
+PREBUILT_DOCKER_REG_PW ?= ""
+PREBUILT_ANAX_VERSION ?= nightly
+PREBUILT_ESS_VERSION ?= nightly
+export PREBUILT_DOCKER_REG_URL
+export PREBUILT_DOCKER_REG_URL
+export PREBUILT_DOCKER_REG_URL
```
linggao, line 64: Why is this exported 3 times?

test/gov/setup_agbot.sh
```diff
@@ -0,0 +1,39 @@
+#!/bin/bash
+echo "setup_agbot.sh start"
+
+echo "Deleting old binaries..."
+rm $ANAX_SOURCE/anax
+rm $ANAX_SOURCE/cli/hzn
```
cgiroua, line 6: Will this script only be run as part of the `test-remote-prebuilt` target? If so, the `clean` should have already removed these

test/Makefile
cgiroua: The exports can also be added to the definition line:
`export PREBUILT_DOCKER_REG_URL ?= ""`

PR #1973, opened 2020-08-19
Issue 989 - Path permissions misaligned

sirotsinskuy:

### DIFF
--- a/cli/dev/utils.go
+++ b/cli/dev/utils.go
@@ -402,6 +402,11 @@ func createContainerWorker() (*container.ContainerWorker, error) {
 		Collaborators: config.Collaborators{},
 	}
 
+	// Create the folder for SSL certificates (under authentication path)
+	if err := os.MkdirAll(config.GetESSSSLClientCertPath(), 0755); err != nil {
+		return nil, err
+	}
+
 	return container.CreateCLIContainerWorker(config)
 }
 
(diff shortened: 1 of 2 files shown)


container/container.go
```diff
@@ -2034,9 +2038,9 @@ func (b *ContainerWorker) ResourcesRemove(agreements []string) error {
 			}
 		}
 
-		// Remove the File Sync Service API authentication credential file.
-		if err := b.GetAuthenticationManager().RemoveCredential(agreementId); err != nil {
-			glog.Errorf("Failed to remove FSS Authentication credential file for %v, error %v", agreementId, err)
+		// Clear the File Sync Service API authentication credential folder
+		if err := os.RemoveAll(b.GetAuthenticationManager().AuthPath); err != nil {
```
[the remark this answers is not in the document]
sirotsinskuy: Thanks for reviewing. So we'll need to remove the UDS file as well, correct? 
Will it be better to remove the "auth" folder along with the UDS file in the end of this function https://github.com/open-horizon/anax/blob/master/cli/dev/utils.go#L675 ?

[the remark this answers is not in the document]
dabooz: Yes, that's a good spot to remove the auth folder as well as the UDS file. Another idea would be to do it here: https://github.com/open-horizon/anax/blob/master/cli/native_deployment/native_deployment.go#L355 by calling a new function that does cleanup, the opposite of what CommonExecutionSetup does. Your choice.

container/container.go
```diff
@@ -248,6 +248,10 @@ func (w *ContainerWorker) finalizeDeployment(agreementId string, deployment *con
 		vols := make(map[string]struct{})
 		for _, bind := range service.Binds {
 			containerVol := strings.Split(bind, ":")
+			// Create all host-paths before docker creates them with root user so we'll be able to clear them without root rights
+			if err := os.MkdirAll(containerVol[0], 0700); err != nil {
```
dabooz, line 252: There are several issues with this part of the change:
1) See https://github.com/open-horizon/anax/blob/master/doc/deployment_string.md. Binds can be files/directories or docker volumes, so you have to account for that here and skip the volume binds.
2) I am concerned that this fix (the 0700 aspect) will break the mac use case. I dont have a mac so I cant test it. The key to testing the mac scenario is getting this script: https://github.com/open-horizon/anax/blob/master/anax-in-container/horizon-container to run the anax-in-container (which you will have to build with your fix in it). Can you test this? If you have trouble, @linggao can help you.
3) I did test this fix on linux, and noticed that hzn dev service stop did NOT clean up the host side files if the service had a file binding in the deployment. I can show you how to set that up if you need help.
4) File bindings can be read-only (ro) and so the file mask on the mkdir needs to reflect that.

Given all the above concerns, this might be enough to consider breaking this fix out into it's own issue so we fix the reported now. Let me know what you think.

sirotsinskuy: Updated the PR

cli/dev/utils.go
```diff
@@ -699,6 +704,20 @@ func stopContainers(dc *common.DeploymentConfig, cw *container.ContainerWorker,
 			cw.ResourcesRemove([]string{msId})
 		}
 	}
+
+	// Remove the UDS socket file
```
dabooz, line 708: Unfortunately, this call happens before all of a service's dependencies have been stopped, so the directory cleanup really needs to be done: https://github.com/open-horizon/anax/blob/master/cli/dev/utils.go#L633 or later.

dabooz (changes requested):

cli/dev/utils.go
sirotsinskuy: Moved cleaning stuff to specified location. Thanks

dabooz (approved):

PR #1977, opened 2020-08-21
Issue 1971 – fix pkg compile for centos/fedora/rhel and aarch64

gavingao-vto:
Signed-off-by: Gavin Gao <gavingao@cn.ibm.com>


losdevnull (commented):
LGTM

Makefile
```diff
@@ -163,6 +163,15 @@ else ifeq ($(opsys_local),Darwin)
 	COMPILE_ARGS_LOCAL += GOOS=darwin
 endif
 
+NO_DEBUG_PKGS := $(shell if [ -f /etc/os-release ];then . /etc/os-release;DISTRO=$$ID;VER=$$VERSION_ID; \
+elif type lsb_release >/dev/null 2>&1;then DISTRO=$$(lsb_release -si);VER=$$(lsb_release -sr); \
+elif [ -f /etc/lsb-release ];then . /etc/lsb-release;DISTRO=$$DISTRIB_ID;VER=$$DISTRIB_RELEASE;fi; \
+if [[ "$$DISTRO" == fedora && $$VER -ge 27 ]];then echo true;elif [[ $$DISTRO =~ ^(centos|rhel)$$ && $$VER -ge 7 ]];then echo true;fi)
+ifeq (${NO_DEBUG_PKGS},true)
+	GO_BUILD_OPTS =-ldflags="-linkmode=external"
+	export RPM_OPTS =--define="%debug_package %{nil}"
```
bmpotter, line 172: The problem with setting RPM_OPTS only in the top level `Makefile` is if someone uses `pkg/rpm/Makefile` or `pkg/rpm/{horizon,horizon-cli}/Makefile` directly it won't be set. I hate to have the complicated code to determine the distro be repeated in all of those Makefiles, so i think the best approach is to put it in a script in the `tools` directory, and then use it in each of the Makefiles.

pkg/rpm/horizon-cli/horizon-cli.aarch64.spec
```diff
@@ -0,0 +1,76 @@
+# Spec file for the horizon-cli RPM. For info about writing spec files, see:
```
bmpotter, line 1: Instead of duplicating the spec files for different architectures, why not just use the `%{_arch}` variable? I would prefer that.

pkg/rpm/horizon/horizon.aarch64.spec
```diff
@@ -0,0 +1,136 @@
+# Spec file for the horizon RPM. For info about writing spec files, see:
```
bmpotter, line 1: Same comment as above.

bmpotter (changes requested):

pkg/rpm/horizon/horizon.aarch64.spec
gavingao-vto: done.

pkg/rpm/horizon-cli/horizon-cli.aarch64.spec
gavingao-vto: done.

Makefile
gavingao-vto: - Makefile in pkg/rpm and pkg/rpm/{horizon,horizon-cli} always call top level Makefile.
```
rpmbuild: require-version
        opsys=Linux arch=$(arch) $(MAKE) -C ../../.. anax
```
- use script "tools/no-debug-pkg" to replace.

gavingao-vto:
@bmpotter Please review.

Makefile
bmpotter: Your comment that "Makefile in pkg/rpm and pkg/rpm/{horizon,horizon-cli} always call top level Makefile" is true for `GO_BUILD_OPTS`, since the target that the lower level makefiles are calling on the top-level Makefile is `anax` and that is where `GO_BUILD_OPTS` is used.

But i think `opsys=Linux arch=$(arch) $(MAKE) -C ../../.. anax` is run in a sub-shell, so i don't think the export of `RPM_OPTS` makes it into the lower-level make processes. Did you test `cd pkg/rpm/horizon && make rpmbuild` and verify that debug_package was set to nil?

bmpotter (changes requested):

Makefile
gavingao-vto: @bmpotter Yes,I had tested these three paths: <pkg/rpm>, <pkg/rpm/horizon>, <pkg/rpm/horizon-cli> in centos8-aarch64/fedora28-aarch64 environments, all of them could make out correct rpm packages.

bmpotter: ok, thx

bmpotter (approved):

Makefile
gavingao-vto: @bmpotter You are right. 
My mistake, in my last test, "export RPM_OPTS" was already in global variables of my shell session.
I've fixed it in PR: https://github.com/open-horizon/anax/pull/2079, and tested again in aarch64 and ppc64le.
Please review.

Issue #1963, opened 2020-08-14
cluster agent failed to start the service pods if the operator is created with helm type


linggao:
The operator was created using 
````
operator-sdk new app-operator --type=helm \
  --api-version=app.example.com/v1alpha1 \  
  --kind=AppService \
  --helm-chart=myrepo/app 
````
The agent was able to create all the objects except the custom resource.
The agent's kube_operator worker cmd process hung while waiting for the customer resource to be created, because it was continually failing.
The operator pod was started but the service pods failed to start, because the operator was not fully initialized when the CR failed to be created.

The root cause is that the operator-sdk doesn't generate the spec.version field in the CRD yaml file. That field is optional, and the agent should compensate for this missing field. The versions list is present and can be used to figure out the correct version of the CR, allowing the CR to be successfully created.

MaxMcAdam:
Verified in 2.27.0-123

PR #1980, opened 2020-08-21
Issue #1963 - cluster agent failed to start the service pods if the o…

MaxMcAdam:
…perator is created with helm type.

The reason is because the operator-sdk doesnt generate the spec.version field in the CRD yaml file. That field is optional, and the agent should compensate for this missing field. The versions list is present and can be used to figure out the correct version of the CR, allowing the CR to be successfully created.

### DIFF
--- a/kube_operator/client.go
+++ b/kube_operator/client.go
@@ -177,7 +177,7 @@ func (c KubeClient) Install(tar string, envVars map[string]string, agId string) 
 		}
 	}
 	// Add the custom resource definitions to the client schema
-	kindToGVRMap := map[string]schema.GroupVersionResource{}
+	kindToGVRMap := map[string][]schema.GroupVersionResource{}
 	for _, crd := range apiObjMap[K8S_CRD_TYPE] {
 		newCRD := crd.Object.(*v1beta1.CustomResourceDefinition)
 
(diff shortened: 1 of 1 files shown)


kube_operator/client.go
```diff
@@ -208,14 +215,26 @@ func (c KubeClient) Install(tar string, envVars map[string]string, agId string)
...
 		unstructCr := unstructured.Unstructured{Object: newCr}
 
 		// the cluster has to create the endpoint for the custom resource, this can take some time
 		glog.V(3).Infof(kwlog(fmt.Sprintf("creating operator custom resource %v", newCr)))
 		for {
-			_, err := crClient.Namespace(namespace).Create(&unstructCr, metav1.CreateOptions{})
+			for _, crClient := range crClients {
+				_, err = crClient.Namespace(namespace).Create(&unstructCr, metav1.CreateOptions{})
+				if err != nil {
+					glog.Errorf("Failed to create custom resource %v.", err)
```
dabooz, line 232: Since this error is likely to happen now that we are looping through clients for every version, let's change to a Warning and add the resource name and version to the message. Maybe we also need a success message for the one that works.

MaxMcAdam: The error is now a warning and there is a sucess message. Also the redundancy if version and versions fields are both specified has been removed

kube_operator/client.go
```diff
@@ -208,14 +214,34 @@ func (c KubeClient) Install(tar string, envVars map[string]string, agId string)
...
+		}
+
 		// the cluster has to create the endpoint for the custom resource, this can take some time
 		glog.V(3).Infof(kwlog(fmt.Sprintf("creating operator custom resource %v", newCr)))
 		for {
-			_, err := crClient.Namespace(namespace).Create(&unstructCr, metav1.CreateOptions{})
+			for version, crClient := range crClients {
+				_, err = crClient.Namespace(namespace).Create(&unstructCr, metav1.CreateOptions{})
+				if err != nil {
+					glog.V(3).Infof(kwlog(fmt.Sprintf("Failed to create custom resource %s with version %v. Error was: %v", resourceName, version, err)))
```
dabooz, line 238: This should be Warningf

MaxMcAdam: Fixed

PR #1984, opened 2020-08-24
Issue 1873 - Agbot reduce the number of calls to the exchange

MaxMcAdam:


exchange/cache.go
```diff
@@ -0,0 +1,286 @@
...
+	}
+	recordHash, err := hashResource(updatedResource)
+	if err != nil {
+		glog.Errorf("Failed to hash resource for cache. Error was : %v", err)
+		recordHash = []byte{}
+	}
+	existingRecord := resourceCache.Get(resourceKey)
+	existingRecordTyped := CacheEntry{}
+	if existingRecord != nil {
+		existingRecordTyped = existingRecord.(CacheEntry)
```
dabooz, line 147: We might need to protect this cast with a check to prevent a panic if the cast fails.

exchange/rpc.go
```diff
@@ -229,6 +236,9 @@ func PutExchangeDevice(httpClientFactory *config.HTTPClientFactory, deviceId str
 			}
 		} else {
 			glog.V(3).Infof(rpclogString(fmt.Sprintf("put device %v to exchange %v", deviceId, pdr)))
+			if cachedNode != nil {
+				UpdateCacheNodePutWriteThru(GetOrg(deviceId), GetId(deviceId), cachedNode, pdr)
```
dabooz, line 240: ok, I see what you're doing here and I understand it. The problem with this approach will occur some time in the future when we add a new field to the node (pdr) and forget to update the UpdateCacheNodePutWriteThru function. Is there a test we could add to the unit tests or something we could add to the code that will catch this problem?

exchange/cache.go
```diff
@@ -0,0 +1,286 @@
...
+	if !ok {
+		return nil
+	}
+	entry := resourceCache.Get(resourceKey)
+	if entry == nil {
+		return nil
+	}
+	typedEntry, ok := entry.(CacheEntry)
+	if !ok {
+		glog.Errorf("Error: object returned from cache not of expected type.")
```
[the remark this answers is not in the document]
MaxMcAdam: Fixed. If the cast fails, the function will return after the error is logged

exchange/cache.go
MaxMcAdam: Fixed

exchange/rpc.go
MaxMcAdam: Couldn't think of a way to do this with a unit test since the test would need to know about a new field in PutDeviceRequest to catch it. Instead, I added a check that all fields in the pdr are used to update the cache. If there is a field in the pdr not used to update the cache, an error will be logged and the cached device deleted

dabooz: Ok, that's a creative solution. Can you change the Infof that logs the message to an Errorf.

MaxMcAdam: Fixed

PR #1994, opened 2020-08-28
Issue 1952 - new format of 'hzn reginput'

sirotsinskuy:

### DIFF
--- a/cli/register/register.go
+++ b/cli/register/register.go
@@ -10,11 +10,13 @@ import (
 	cliexchange "github.com/open-horizon/anax/cli/exchange"
 	"github.com/open-horizon/anax/cli/unregister"
 	"github.com/open-horizon/anax/common"
+	"github.com/open-horizon/anax/compcheck"
 	"github.com/open-horizon/anax/cutil"
 	"github.com/open-horizon/anax/exchange"
 	"github.com/open-horizon/anax/externalpolicy"
 	"github.com/open-horizon/anax/i18n"
 	"github.com/open-horizon/anax/persistence"
+	"github.com/open-horizon/anax/policy"
 	"github.com/open-horizon/anax/semanticversion"
 	"io/ioutil"
 	"k8s.io/client-go/rest"
@@ -25,11 +27,6 @@ import (
 	"strings"
 	"time"
 )
-
-type ExchangeNodes struct {
-	LastIndex int                        `json:"lastIndex"`
-	Nodes     map[string]exchange.Device `json:"nodes"`
-}
 
 // read the user input file that contains the global attributes and service userinput settings.
 // it supports both old and new format.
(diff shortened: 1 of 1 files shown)


cli/register/register.go
```diff
@@ -777,10 +779,23 @@ func AddAllRequiredSvcs(nodeCreds, org, url, arch, versionRange string, allRequi
 	allRequiredSvcs[svcKey].HighestVersion = highestSvc.Version // in case we don't encounter this service again, we already know the highest version for getting the user input from
 	allRequiredSvcs[svcKey].UserInputs = highestSvc.UserInputs
 
+	// check if the service contains privileged
```
linggao, line 782: Please use exiting function to check https://github.com/open-horizon/anax/blob/63f6076e82b2d203a1d7bbb222644058ef95c944/compcheck/comp_check.go#L543

sirotsinskuy: Didn't know about that function, thanks. But we need to somehow mark result of `AddAllRequiredSvcs()` function that some of services contain privileged. So the best way I found - is marking each service separately (https://github.com/open-horizon/anax/pull/1994/files#diff-09df09ce2e6c43d6445e8d0c86f8c681R790) so that services can be checked later (https://github.com/open-horizon/anax/pull/1994/files#diff-09df09ce2e6c43d6445e8d0c86f8c681R862) . And with that function we can check if deployment contains privileged, but then we'll need to pass that result to calling function. 
Another option is to make `AddAllRequiredSvcs()` return boolean with result of `DeploymentRequiresPrivilege()`, but the name of function in this case doesn't represent that it's returning smth, and we'll have strange `if AddAllRequiredSvcs() ... {}` block.
Using of `ServicesRequirePrivilege` doesn't help as well, as we need to mark each service separately (with current implementation)
Do you have any suggestions how that could be implemented better? Maybe there is some way to get the deployment string in `CreateInputFile` itself (I can't find the possibilities now) ?

linggao: Sorry, I do not quite follow you. Can you just replace line 783-794 with the following code?
````
if priv, err := compcheck.DeploymentRequiresPrivilege(highestSvc.Deployment, msgPrint); err != nil {
     ....
} elsie if priv {
    allRequiredSvcs[svcKey].Privileged = true
}
````

Comment (sirotsinskuy) [reply cli/register/register.go:782]: Sorry, didn't get your request right. Updated PR

**Review by linggao — CHANGES_REQUESTED** [review]
Need to add check for node type.

File: cli/register/register.go [hunk 2]
```diff
@@ -697,13 +694,33 @@ type SvcMapValue struct {
...
+
+	var nodes exchange.GetDevicesResponse
+	cliutils.ExchangeGet("Exchange", cliutils.GetExchangeUrl(), "orgs/"+nodeOrg+"/nodes/"+nodeId, nodeCreds, []int{200}, &nodes)
+
+	// Make sure that the node type and the service type match
+	serviceType := highestSvc.GetServiceType()
+	for nId, n := range nodes.Devices {
+		if serviceType != exchange.SERVICE_TYPE_BOTH && n.GetNodeType() != serviceType {
+			msgPrinter.Printf("Ignoring version %s of service %s/%s with node type mismatch: the service type '%v' does not match the node type '%v' of the Exchange node %v.", versionRange, org, url, serviceType, n.GetNodeType(), nId)
+			msgPrinter.Println()
```
linggao, line 719: Please change the above 2 lines to `cliutils.Verbose(msgPrinter.Sprintf(....)`

linggao (approved):

Issue #1862, opened 2020-06-11
Cache service keys and dockauths


MaxMcAdam:
Starting with registration, agreement negotiation, and any process that is likely to affect many nodes simultaneously

MaxMcAdam:
Calls made to the exchange during registration with stack traces[https://app.zenhub.com/files/70207795/f792a0ec-206a-4095-8ab1-2eb109bb951c/download](https://app.zenhub.com/files/70207795/f792a0ec-206a-4095-8ab1-2eb109bb951c/download)

dlarson04:
Here is something that could possibly be improved with the messages... Here is the exchange logs showing calls from an agent to the exchange

```
[06/30/2020 18:56:31.148] icp-console.apps.dominion.ieam-vsphere.com:dominion/edge-scale-test250_horizon0 GET http://icp-console.apps.dominion.ieam-vsphere.com:443/v1/orgs/dominion/nodes/edge-scale-test250_horizon0/msgs: 404 Not Found
[06/30/2020 18:56:47.983] ibm-edge-exchange:IBM/dominion-agbot POST http://ibm-edge-exchange/v1/orgs/dominion/nodes/edge-scale-test250_horizon0/msgs: 201 Created
[06/30/2020 18:56:48.005] ibm-edge-exchange:IBM/dominion-agbot POST http://ibm-edge-exchange/v1/orgs/dominion/nodes/edge-scale-test250_horizon0/msgs: 201 Created
[06/30/2020 18:56:52.498] icp-console.apps.dominion.ieam-vsphere.com:dominion/edge-scale-test250_horizon0 GET http://icp-console.apps.dominion.ieam-vsphere.com:443/v1/orgs/dominion/nodes/edge-scale-test250_horizon0/msgs: 200 OK
[06/30/2020 18:56:52.700] icp-console.apps.dominion.ieam-vsphere.com:dominion/edge-scale-test250_horizon0 GET http://icp-console.apps.dominion.ieam-vsphere.com:443/v1/orgs/dominion/nodes/edge-scale-test250_horizon0/msgs: 200 OK
[06/30/2020 18:56:52.703] icp-console.apps.dominion.ieam-vsphere.com:dominion/edge-scale-test250_horizon0 GET http://icp-console.apps.dominion.ieam-vsphere.com:443/v1/orgs/dominion/nodes/edge-scale-test250_horizon0/msgs: 200 OK
[...]

PR #2012, opened 2020-09-08
Issue #1862 - Cache service keys and dockauths

MaxMcAdam:


exchange/cache.go
```diff
@@ -80,6 +82,26 @@ func GetServicePolicyFromCache(svcOrg string, svcId string, svcArch string, svcV
 	return nil
 }
 
+// GetServicePolicyFromCache returns the service policy from the exchange cache if it is present, or nil if it is not
```
dabooz, line 85: copy/paste problem in the comments, referencing service policy.

dabooz (changes requested):

exchange/cache.go
MaxMcAdam: Fixed

dabooz (approved):

PR #2015, opened 2020-09-09
Issue 1694: overhaul of agent-install.sh

bmpotter:
Signed-off-by: Bruce Potter <bp@us.ibm.com>

### DIFF
--- a/agent-install/agent-install.sh
+++ b/agent-install/agent-install.sh
@@ -1,32 +1,31 @@
 #!/bin/bash
 
-# The script installs Horizon agent on an edge node
+# Installs the Horizon agent on an edge node (device or edge cluster)
 
-set -e
+set -e   #future: remove?
 
+#====================== Input Gathering ======================
 
-SCRIPT_VERSION="1.1.1"
+# Global constants
 
-SUPPORTED_OS=( "macos" "linux" )
-SUPPORTED_LINUX_DISTRO=( "ubuntu" "raspbian" "debian" )
-SUPPORTED_LINUX_VERSION=( "bionic" "buster" "xenial" "stretch" )
-SUPPORTED_ARCH=( "amd64" "arm64" "armhf" )
+# logging levels
+#VERB_FATAL=0   # we always show fata error msgs
+VERB_ERROR=1
+VERB_WARNING=2
+VERB_INFO=3
+VERB_VERBOSE=4
+VERB_DEBUG=5
 
-# Defaults
-PKG_PATH="."
-PKG_TREE_IGNORE=false
-SKIP_REGISTRATION=false
-CFG="agent-install.cfg"
-OVERWRITE=false
-SKIP_PROMPT=false
-HZN_NODE_POLICY=""
-AGENT_INSTALL_ZIP="agent-install-files.tar.gz"
-NODE_ID_MAPPING_FILE="node-id-mapping.csv"
-CERTIFICATE_DEFAULT="agent-install.crt"
-BATCH_INSTALL=0
-DEPLOY_TYPE="device" # "cluster" for deploy on edge cluster
+SUPPORTED_OS=("macos" "linux")
+SUPPORTED_LINUX_DISTRO=("ubuntu" "raspbian" "debian")
+SUPPORTED_LINUX_VERSION=("bionic" "buster" "xenial" "stretch")
+SUPPORTED_ARCH=("amd64" "arm64" "armhf")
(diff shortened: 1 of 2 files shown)


agent-install/agent-install.sh
```diff
@@ -35,1269 +34,1048 @@ CONFIGMAP_NAME="openhorizon-agent-config"
...
+        get_variable APT_REPO_BRANCH 'updates'
+    elif [[ "${AGENT_DEPLOY_TYPE}" == "cluster" ]]; then
+        get_variable EDGE_CLUSTER_STORAGE_CLASS 'gp2'
+        get_variable AGENT_NAMESPACE 'openhorizon-agent'
+        get_variable USE_EDGE_CLUSTER_REGISTRY 'true'
+        get_variable AGENT_DEPLOYMENT_STATUS_TIMEOUT_SECONDS '75'
+
+        if [[ "$USE_EDGE_CLUSTER_REGISTRY" == "true" ]]; then
+            #lily: do we really need the user to input IMAGE_ON_EDGE_CLUSTER_REGISTRY and AGENT_IMAGE_TAG. These seem like internal values (our code creates the image in the local registry with these values and then uses that image). Can we just always default them based on the edge cluster type?
+            get_variable IMAGE_ON_EDGE_CLUSTER_REGISTRY '' 'true'
```
LiilyZhang, line 328: `IMAGE_ON_EDGE_CLUSTER_REGISTRY`:
OCP - `<registry-host>/<agent-project>/amd64_anax_k8s`
microk8s - `localhost:32000/<agent-namespace>/amd64_anax_k8s`
k3s - `<docker-registry-service-endpoint>/openhorizon-agent/agent_amd64`

So if we want to default `IMAGE_ON_EDGE_CLUSTER_REGISTRY` for different types, we might need to retrieve edge OCP `<registry-host>` on the fly. Also we need to get the `<docker-registry-service-endpoint>` for k3s

LiilyZhang: The `AGENT_IMAGE_TAG` is actually put into `agent-install.cfg` file by `edgeNodeFiles.sh`. The value is passed in to `edgeNodeFiles.sh` by specifying `-i` (if not, use default image tag). So it is info from the agent image. 
But I think it won't hurt anything if we remove this value from cfg file, and variable check in `agent-install.sh`. If the agent image tag is removed, prob we need to parse the image name and extract the version in install script?

bmpotter: Yes, that's what i meant, have the code figure out what the values should be

agent-install/agent-install.sh
```diff
@@ -1670,29 +1357,30 @@ function cleanup_cluster_config_files() {
...
+                image_full_path_on_edge_cluster_registry_internal_url=$DEFAULT_OCP_INTERNAL_URL_FOR_EDGE_CLUSTER_REGISTRY/$EDGE_CLUSTER_REGISTRY_PROJECT_NAME/$EDGE_CLUSTER_AGENT_IMAGE_AND_TAG
+            fi
+        else
+            image_full_path_on_edge_cluster_registry_internal_url=$INTERNAL_URL_FOR_EDGE_CLUSTER_REGISTRY/$EDGE_CLUSTER_REGISTRY_PROJECT_NAME/$EDGE_CLUSTER_AGENT_IMAGE_AND_TAG
+        fi
+        sed -i -e "s#__ImagePath__#${image_full_path_on_edge_cluster_registry_internal_url}#g" deployment.yml
     else
-		log_notify "Agent install on edge cluster requires to use an edge cluster registry"
-		exit 1
+        #lily: this makes it look like the only acceptable value for USE_EDGE_CLUSTER_REGISTRY is 'true'. If that is the case, why do we have this variable?
```
LiilyZhang, line 1382: So when started to implement agent install script for cluster install during last release, we had mentioned 2 ways for getting/storing agent image: 1) edge cluster image registry, 2) public registry. But later on we focused on edge cluster image registry and didn't talk about public registry anymore (also didn't mention use public registry in our doc), so the code is like only implemented the path that use the edge cluster image registry. 
So do we want to support public registry? I think we can get rid of `USE_EDGE_CLUSTER_REGISTRY` if we only want to support edge cluster that with image registry

bmpotter: Thx, the background helps. I think we **do** want to support a public registry eventually. So we should keep the variable internally, but for now remove it from usage documentation. I can do that.

Issue #1967, opened 2020-08-17
hzn service log could not display with not existing path: /var/log/syslog


georgelee-foxconn:
I had tried to install rpm of horizon and horizon client, and do register to management hub with sample hellowolrd service.
I would like to use the command to check whether the service is running or not. Here is the command and the response error:

Request: 
hzn service log -f ibm.helloworld

Response: 
Displaying log messages for service ibm.helloworld with service id 124b8714b56510d80f94aa5cbacbd1a2a0f2cfa0163a086df1d4771a02ad6a3e.
Use ctrl-C to terminate this command.
Error: /var/log/syslog could not be opened or does not exist: open /var/log/syslog: no such file or directory

1 I've tried in two different edge environments, x86 RHEL and arm64 CentOS.,both have the same response.
   And both environments do not have the path: /var/log/syslog
2. horizon-cli is built-up by project: anax/pkg/rpm.
3. ibm.helloworld is running and confirmed by command: docker ps

georgelee-foxconn:
From the message in following link: https://github.com/open-horizon/anax/tree/master/pkg/rpm

The hzn service log command does not work correctly, so view the edge service output like this:
tail -f /var/log/messages | grep ibm.helloworld

It could be worked to check the status from /var/log/messages

losdevnull:
So when agent launchs docker containers it always specifies `syslog` as docker log driver type.  However different linux dist has different default log file path for syslog.
We could enhance `hzn service log` to recoginize different linux dist and refer to the correct syslog file for container logs.

PR #2025, opened 2020-09-10
Issue 1967 - Determine system log file path based on linux distribution

losdevnull:
Signed-off-by: Liang Wang <losdevnull@gmail.com>

issue #1967

### DIFF
--- a/cli/service/service.go
+++ b/cli/service/service.go
@@ -169,10 +169,19 @@ func LogMac(instanceId string, tailing bool) {
 
 func LogLinux(instanceId string, tailing bool) {
 	msgPrinter := i18n.GetMessagePrinter()
+	// Determine the system log file based on linux distribution
+	sysLogPath := "/var/log/messages"
+	_, err := os.Stat("/etc/redhat-release")
+	if os.IsNotExist(err) {
+		_, err := os.Stat("/etc/centos-release")
+		if os.IsNotExist(err) {
+			sysLogPath = "/var/log/syslog"
+		}
+	}
 	// The requested service is running, so grab the records from syslog for this service.
-	file, err := os.Open("/var/log/syslog")
+	file, err := os.Open(sysLogPath)
 	if err != nil {
-		cliutils.Fatal(cliutils.NOT_FOUND, msgPrinter.Sprintf("/var/log/syslog could not be opened or does not exist: %v", err))
+		cliutils.Fatal(cliutils.NOT_FOUND, msgPrinter.Sprintf("%v could not be opened or does not exist: %v", sysLogPath, err))
 	}
 	defer file.Close()
 	// Check file stats and capture the current size of the file if we will be tailing it.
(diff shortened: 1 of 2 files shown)


cli/locales/de/messages.gotext.json
```diff
@@ -15917,64 +15917,96 @@
       "fuzzy": true
     },
     {
-      "id": "/var/log/syslog could not be opened or does not exist: {Err}",
-      "message": "/var/log/syslog could not be opened or does not exist: {Err}",
-      "translation": "'/var/log/syslog' konnte nicht geöffnet werden oder ist nicht vorhanden, {Err}",
+      "id": "{SysLogPath} could not be opened or does not exist: {Err}",
```
dabooz, line 15920: These files are generated by the go text tools, please do not change any files that end in gotext*.json. The files with non-English translations are also tool generated, IBM donates them to this project. Please remove them all from this PR.

cli/service/service.go
```diff
@@ -4,6 +4,14 @@ import (
 	"bufio"
 	"encoding/json"
 	"fmt"
+	"io"
```
dabooz, line 7: Please use `make format` to format source code so that these kinds of changes don't happen and the imports will remain in alphabetic order.

cli/service/service.go
```diff
@@ -169,18 +170,27 @@ func LogMac(instanceId string, tailing bool) {
...
+		if os.IsNotExist(err) {
+			sysLogPath = "/var/log/syslog"
+		}
+	}
 	// The requested service is running, so grab the records from syslog for this service.
-	file, err := os.Open("/var/log/syslog")
+	file, err := os.Open(sysLogPath)
 	if err != nil {
-		cliutils.Fatal(cliutils.NOT_FOUND, msgPrinter.Sprintf("/var/log/syslog could not be opened or does not exist: %v", err))
+		cliutils.Fatal(cliutils.NOT_FOUND, msgPrinter.Sprintf("%v could not be opened or does not exist: %v", sysLogPath, err))
```
dabooz, line 185: The go text tools will pick up this message change and will automatically adjust the gotext files.

dabooz (changes requested):

cli/locales/de/messages.gotext.json
losdevnull: Removed all changes to gotext files

cli/service/service.go
losdevnull: I tried make format but it did not restore the imports back to alphabetic order.   Thus I reverted it manually.   
btw, the original order change was done automatically by VSCode golang source format, I guess it was trying to organize the imports so that default packages would be on top.

cli/service/service.go
losdevnull: Thanks @dabooz!  It is good to know that gotext translation are handled by tools via IBM donation.   I was actually wondering how that would be handled, until I saw an example like this:  https://github.com/open-horizon/anax/pull/2031
I've reverted all changes to gotext*.json files.

losdevnull: I also further updated pkg/rpm/README to remove the limitation statement.

dabooz (approved):

Issue #2016, opened 2020-09-09
anax should invalidate the cache when an org is removed


linggao:
Anax should invalidate the cache for all the objects that are related to an org when the org is removed.
Please work with the exchange team to make sure the exchange put this event in the changes list.

MaxMcAdam:
Verified in e2edev

PR #2032, opened 2020-09-10
Issue #2016 - anax should invalidate the cache when an org is removed

MaxMcAdam:

### DIFF
--- a/exchange/cache.go
+++ b/exchange/cache.go
@@ -219,6 +219,26 @@ func DeleteCacheResource(resourceType string, resourceKey string) interface{} {
 	return retResource
 }
 
+// DeleteOrgCachedResources will delete all cached resources from the given org
+func DeleteOrgCachedResources(org string) {
+	glog.V(5).Infof("Delete all resources from org %v", org)
+	if ExchangeResourceCache == nil || ExchangeResourceCache.allResources == nil {
+		return
+	}
+
+	ExchangeResourceCache.Lock.Lock()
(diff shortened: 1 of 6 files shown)


MaxMcAdam:
The cache will drop the resources for an org when the agent gets a change message for the org with operation created or deleted. The created/modified operation means that there was a PUT request to the org in the exchange. This does not effect the cache because a new org cannot be created by a PUT request and the fields that can be changed by a PUT (or a PATCH) are not in the exchange cache.

exchange/cache.go
```diff
@@ -309,6 +329,8 @@ func DeleteCacheResourceFromChange(change ExchangeChange, nodeId string) {
 	} else if change.IsServicePolicy() {
 		id, arch, vers := svcInformationFromSvcId(change.ID)
 		DeleteCacheResource(SVC_POL_TYPE_CACHE, ServicePolicyCacheMapKey(change.OrgID, id, arch, vers))
+	} else if change.IsOrg() && (change.Operation == CHANGE_OPERATION_CREATED || change.Operation == CHANGE_OPERATION_DELETED) {
```
dabooz, line 332: There are actually 4 events which can occur where everything in the org cache needs to be deleted. You handle 2 of them here. The other 2 are when the org is added to or removed from the agbot's served list. See this code: https://github.com/open-horizon/anax/blob/master/agreementbot/changes_worker.go#L270

dabooz (changes requested):

dabooz (approved):

Issue #1926, opened 2020-07-24
Remove horizon.pem and mtn-publicKey.pem from deb and RPM packages


dabooz:
Those certs should no longer be needed, but this needs to be tested.
And, we need to make sure that e2edev still runs without them (or it's updated to run without them).

IntelliJIN:
WIP

bmpotter:
Note that 1 of the reasons these files were left in the `horizon` pkg was because we moved them from the `bluehorizon` pkg to `horizon`, and when they weren't in the `horizon` pkg, `apt-get install ...` didn't completely git rid of the `bluehorizon` pkg. So while you are making this change, you need to test the case it which the system has an older `bluehorizon`, `horizon`, and `horizon-cli` deb pkg installed and you upgrade to your new `horizon` and `horizon-cli` pkgs (and run dpkg -l to see what is left).

Also, the files should also be removed from the `horizon` rpm. (For the rpms you don't have the same migration issue, because there has never been a `bluehorizon` rpm.

bmpotter:
@IntelliJIN I'm reviewing https://github.com/open-horizon/anax/pull/2037, but you never responded to my comment above, so i don't know if you addressed and tested my concern about upgrading on a system that had the `bluehorizon` deb pkg installed.

bmpotter:
Haven't heard from @IntelliJIN in a while. We might have to take this issue over and complete the in-progress PR https://github.com/open-horizon/anax/pull/2037 .

rajpratyush:
@bmpotter if you guide me I would like to complete the PR

bmpotter:
@rajpratyush Yes, you can work on this! It looks like @IntelliJIN doesn't have time to complete it. Look at his [PR}(https://github.com/open-horizon/anax/pull/2037) and start from there. But notice my comment above in this issue.

rajpratyush:
@bmpotter sure sir i am on it

dabooz:
@rajpratyush I updated the labels on this issue to reflect that it is in progress again.

PR #2037, opened 2020-09-11
Issue/1926 remove horizon.pem and mtn public key.pem from deb and rpm packages

IntelliJIN:
Key and cert are removed from the repository and are not using in new deb and rpm builds.
There are no specific changes to remove cert and key we can still have in fs after previous installations.
Config CACertsPath now pointing not to specific cert, but to dir with a single or multiple cert files. 
Added support for DER encoded certs. List of supported extensions - .pem, .der, .cer and .crt ( both PEM and DER-encoded )


dabooz (commented):
The e2edev test environment also needs to be updated here: https://github.com/open-horizon/anax/blob/master/test/Makefile#L279 to stop putting these old keys and certs into the agent configuration. Removing this code will test that everything works without the old keys and certs.

bmpotter (changes requested):
I need to know whether you addressed my comment in https://github.com/open-horizon/anax/issues/1926

PR #2042, opened 2020-09-14
Issue 1970 - move doc to docs

dabooz:
Signed-off-by: Dave Booz <booz@us.ibm.com>


joewxboy (approved):
I believe that's correct.  Once I create the Jekyll config directories, we could then see about adding any front matter to the markdown files.  But this should set us up nicely for that change.

Issue #2021, opened 2020-09-09
Finalizer on an operator custom resource prevents the cr and crd from being deleted


MaxMcAdam:
Need to make sure the cr is truly deleted before attempting to uninstall any other components

MaxMcAdam:
Verified in local build

PR #2068, opened 2020-09-16
Issue #2021 - Finalizer on an operator custom resource prevents the c…

MaxMcAdam:
…r and crd from being deleted

### DIFF
--- a/kube_operator/api_objects.go
+++ b/kube_operator/api_objects.go
@@ -43,8 +43,8 @@ func sortAPIObjects(allObjects []APIObjects, customResource *unstructured.Unstru
 				} else {
 					return objMap, namespace, fmt.Errorf(kwlog(fmt.Sprintf("Error: namespace object must have a name in its metadata section.")))
 				}
-				if namespace != typedNS.ObjectMeta.Name {
-					return objMap, namespace, fmt.Errorf(kwlog(fmt.Sprintf("Error: multiple namespaces specified in operator: %s and %s", namespace, typedNS.ObjectMeta.Name)))
+				if namespace != "" && namespace != typedNS.ObjectMeta.Name {
+					return objMap, namespace, fmt.Errorf(kwlog(fmt.Sprintf("Error: multiple namespaces specified in operator : %s and %s", namespace, typedNS.ObjectMeta.Name)))
 				}
 				namespace = typedNS.ObjectMeta.Name
 			} else {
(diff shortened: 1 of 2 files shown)


MaxMcAdam:
The deployment object handles the execution of finalizers. If the operator custom resource has a finalizer and the deployment object is uninstalled before the finalizer can finish executing, the finalizer will not be removed from the CR and the CR becomes unkillable until you manually patch the CR to remove the finalizer. This also blocks the CRD from being removed successfully.

kube_operator/api_objects.go
```diff
@@ -456,6 +441,27 @@ func (cr CustomResourceV1Beta1) Uninstall(c KubeClient, namespace string) {
...
 	if err != nil {
 		glog.Warningf(kwlog(fmt.Sprintf("unable to delete operator custom resource %s. Error: %v", newCrName, err)))
+	} else {
+		status, err := cr.Status(c, namespace)
+		timeout := 20
+		for (err == nil || status != nil) && timeout > 0 {
+			glog.Infof(kwlog(fmt.Sprintf("Custom Resource %s is not yet down. Pausing for 10 before checking again. Custom resource status is: %v Error is %v", newCrName, status, err)))
+			time.Sleep(10 * time.Second)
+			timeout = timeout - 1
+		}
```
dabooz, line 451: Put an error message in the log if the timeout occurs, and in the same code below too. You could make a reusable function for the whole status check too.

kube_operator/client.go
```diff
@@ -333,7 +337,22 @@ func getK8sObjectFromYaml(yamlFiles []YamlFile, sch *runtime.Scheme) ([]APIObjec
 	_ = v1scheme.AddToScheme(sch)
 	_ = scheme.AddToScheme(sch)
 
-	for _, fileStr := range yamlFiles {
+	// multiple yaml files can be in one file separated by '---'
+	// these are split here and rejoined with the single files
```
dabooz, line 341: i really dislike yaml...the person who thought this was a good idea should have their coding license revoked.

dabooz (changes requested):

kube_operator/api_objects.go
MaxMcAdam: Fixed in both

kube_operator/client.go
MaxMcAdam: can't trust anything that puts so much syntactic significance on whitespace

dabooz (approved):

PR #2071, opened 2020-09-17
Issue 2063 - On agent registering, if call to maxchangeid fails chang…

linggao:
…esId never gets set so changes route is never called

### DIFF
--- a/changes/changes_worker.go
+++ b/changes/changes_worker.go
@@ -120,20 +120,9 @@ func (w *ChangesWorker) Initialize() bool {
 	// will be non-zero. This logic is placed here (and not the constructor) because we dont want to emit messages
 	// to the message bus from inside the worker constructor.
 	if w.changeID == 0 && w.GetExchangeToken() != "" {
-
-		// Call the exchange to retrieve the current max change id. Use a custom exchange context that blocks (retries forever)
-		// until we can get the max change ID.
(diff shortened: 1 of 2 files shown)


changes/changes_worker.go
```diff
@@ -120,26 +120,9 @@ func (w *ChangesWorker) Initialize() bool {
...
-		// until we can get the max change ID.
-		ec := exchange.NewCustomExchangeContext(w.EC.Id, w.EC.Token, w.EC.URL, w.EC.CSSURL, w.Config.Collaborators.HTTPClientFactory)
-		if maxChangeID, err := exchange.GetHTTPExchangeMaxChangeIDHandler(ec)(); err != nil {
-			glog.Errorf(chglog(fmt.Sprintf("Error retrieving max change ID, error: %v", err)))
-		} else {
-			w.changeID = maxChangeID.MaxChangeID
-			if err := persistence.SaveExchangeChangeState(w.db, w.changeID); err != nil {
-				glog.Errorf(chglog(fmt.Sprintf("error saving persistent exchange change state, error %v", err)))
-			}
+		if err := w.getChangeId(); err != nil {
```
dabooz, line 123: This change will break because it does not set the polling intervals correctly if anax is restarted after it is registered, see old lines 140-142. That's why they are here, outside the if statement on old line 122.

dabooz (changes requested):

dabooz (approved):

PR #2074, opened 2020-09-17
Issue 2026 - Display or log warning when node reach 5% org maxNodes Limits in CLI

IntelliJIN:
Issue 2026 - Display or log warning when node reach 5% org maxNodes Limits in CLI

Signed-off-by: Igor Kolinko <ikolin@softserveinc.com>

### DIFF
--- a/cli/exchange/node.go
+++ b/cli/exchange/node.go
@@ -135,6 +135,10 @@ func NodeCreate(org, nodeIdTok, node, token, userPw, arch string, nodeName strin
 	}
 
 	var httpCode int
+	var resp struct {
+		Code string `json:"code"`
+		Msg  string `json:"msg"`
+	}
 	if nodeExists {
 		msgPrinter.Printf("Node %v exists. Only the node token will be updated.", nodeId)
 		msgPrinter.Println()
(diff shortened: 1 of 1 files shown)


cli/exchange/node.go
```diff
@@ -180,6 +185,10 @@ func NodeCreate(org, nodeIdTok, node, token, userPw, arch string, nodeName strin
 		} else {
 			msgPrinter.Printf("Node %v created.", nodeId)
 			msgPrinter.Println()
+			if strings.Contains(resp.Msg, "Warning: you are within") {
+				msgPrinter.Printf("Note: the org %v has reached to the 95%% of the node limit.", org)
```
dabooz, line 189: Please change the text to "Note: the org %v has reached 95%% of the node limit: %v.", org, nodeLimit
I think you will have to call GET on the exchange  /orgs/{org} resource to get the limit, if the limit is not in the exchange message.

dabooz (changes requested):

cli/exchange/node.go
linggao: Do not worry about the limits here, the exchange-api will change the text. It will have the limit in the message.

cli/exchange/node.go
```diff
@@ -180,6 +185,10 @@ func NodeCreate(org, nodeIdTok, node, token, userPw, arch string, nodeName strin
 		} else {
 			msgPrinter.Printf("Node %v created.", nodeId)
 			msgPrinter.Println()
+			if strings.Contains(resp.Msg, "Warning: you are within") {
```
linggao, line 188: The message gotten from the exchange might be translated if the exchange has a non-English locale set. Comparing  the text here will have problem in this case.

cli/exchange/node.go
IntelliJIN: What is the new message from exchange-api? Maybe it will help to resolve both issues.

linggao (approved):

dabooz (approved):

PR #2098, opened 2020-09-23
Issue 2096 - Fix no-debug-pkg versions issue

### LINKED ISSUE #2096: no-debug-pkg failed when building at RHEL7.x-8.x
Output of `cat /etc/os-release` at RHEL distros has fractional format of distro version. This leads to the following bug when I try to build anax RPM packages on RHEL machine (7.8 in this case):
```
no-debug-pkg: line 17: [[: 7.8: syntax error: invalid arithmetic operator (error token is ".8")
```
The same behavior is in RHEL 8.2.

Expected that no-debug-pkg will ignore fractional part of OS version and provide needed flags to the build environment by returning true if the following condition matched:
```bash
elif [[ $DISTRO =~ ^(centos|rhel)$ && $VER -ge 7 ]];then
```

To reproduce, for RPM builds as an example, run `make rpmpkgs`.

P.S. Fedora/CentOS releases are ok - major release version ID mentioned there.

jacob-rogers:
This PR fixes version ID matching at no-debug-pkg script. Affects external linking mode switch on RHEL platform builds, that omits debuginfo packages and the corresponding rpmbuild breakdown on RHEL.

Signed-off-by: Dmitry Gorbachev <dmitry.gorbachev@ru.ibm.com>

Resolves #2096

### DIFF
--- a/tools/no-debug-pkg
+++ b/tools/no-debug-pkg
@@ -12,6 +12,9 @@ elif [ -f /etc/lsb-release ];then
 	DISTRO=$DISTRIB_ID;VER=$DISTRIB_RELEASE;
 fi
 
+# get rid of version's fraction (major releases make sense)
+VER=${VER%%.*}
+
 if [[ "$DISTRO" == fedora && $VER -ge 27 ]];then
 	echo true;
 elif [[ $DISTRO =~ ^(centos|rhel)$ && $VER -ge 7 ]];then


jacob-rogers:
@dabooz This PR is ready, pls take a look - fixing a build time issue

bmpotter (approved):

PR #2111, opened 2020-09-25
Issue 1510 - Multiple Agents

nhjabbour:


test/Makefile
```diff
@@ -103,26 +106,32 @@ run-agbot: agbot-docker-prereqs test-network
 		--add-host "$(DOCKER_AGBOT_ADD_HOST)" \
 		-v /var/run/docker.sock:/var/run/docker.sock \
 		-v $(UDS):$(UDS):rw \
+		-v /var/tmp/horizon/:/var/tmp/horizon/:rw \
```
dabooz, line 109: This needs to be /var/tmp/horizon on the host machine?

test/gov/init_exchange.sh
```diff
@@ -207,6 +207,26 @@ echo "Registering Anax device1 in customer org..."
...
 echo "$REGANAX1C"
 
+DEVICE_NUM=6
+if [ ${MUL_AGENTS} -lt 5 ]; then
+  NUM_AGENTS=$((${MUL_AGENTS}+$DEVICE_NUM))
+else
+  NUM_AGENTS=$((4+$DEVICE_NUM))
+fi
+while [ ${DEVICE_NUM} -lt ${NUM_AGENTS} ]; do
+  echo "Registering Anax device${DEVICE_NUM}..."
```
dabooz, line 217: Do we have to register the external agents this way? is there no way we could use hzn register for them?

test/gov/verify_agreements.sh
```diff
@@ -9,8 +9,43 @@ echo -e "${PREFIX} starting"
...
+        MUL_NUM_AG_1=0
+        MUL_NUM_AG_2=0
+        MUL_NUM_AG_3=0
+        MUL_NUM_AG_4=0
         if [ "${HZN_REG_TEST}" != "1" ]; then
             if [ "${PATTERN}" == "sall" ]; then
+              case ${MUL_AGENTS} in
+                1)
+                echo -e "${PREFIX} 1 extra agent"
+                MUL_NUM_AG_1=6
```
dabooz, line 21: Did you enumerate the totals this way so that each agent could have a different config and therefore a different number of agreements? otherwise this is really un-necessary

test/gov/run_agent.sh
```diff
@@ -0,0 +1,28 @@
+source /tmp/exports_list
+EX_IP_GATEWAY=$(docker inspect exchange-api | jq -r '.[].NetworkSettings.Networks.e2edev_test_network.Gateway')
+EX_HOST_PORT=$(docker inspect exchange-api | jq -r '.[].NetworkSettings.Ports."8080/tcp"[].HostPort')
+export HZN_EXCHANGE_URL="http://$EX_IP_GATEWAY:$EX_HOST_PORT/v1"
+export EXCH_APP_HOST="http://$EX_IP_GATEWAY:$EX_HOST_PORT/v1"
+AGENT_PORT=$(docker port "$(cat /proc/self/cgroup | grep "docker" | sed s/\\//\\n/g | tail -1)" | awk -F'/' '{print $1}')
+echo ${AGENT_PORT}
+DEVICE_NUM=$((${AGENT_PORT} - 8512 + 6))
+echo ${DEVICE_NUM}
+export USER=anax1
```
dabooz, line 10: Did you really need to export all of these env vars given that your already sourced many of them on line 1?

test/gov/run_agent.sh
```diff
@@ -0,0 +1,28 @@
...
+export ANAX_API="http://localhost:${HZN_AGENT_PORT}"
+export EXCH=$HZN_EXCHANGE_URL
+
+export PATTERN="sall"
+export NOK8S=1
+export constraint2="NOK8S == true"
+export PARTNERID="$DEVICE_ORG/anaxdevice${DEVICE_NUM}"
+export HZN_MGMT_HUB_CERT_PATH="/tmp/css.crt"
+
+./apireg.sh > agent_debug.log
```
dabooz, line 28: I think it would be better to use hzn register on these agents, so that we test that path.

test/gov/multiple_agents.sh
```diff
@@ -0,0 +1,38 @@
...
+  export HZN_EXCHANGE_URL="http://$EX_IP_GATEWAY:$EX_HOST_PORT/v1";
+  export HZN_EXCHANGE_USER_AUTH="e2edev@somecomp.com/e2edevadmin:e2edevadminpw";
+  export HZN_FSS_CSSURL=${CSS_URL};
+  export HZN_AIC="/tmp/anax-in-container/horizon-container";
+  export HORIZON_NUM=${DEVICE_NUM};
+  #export HC_DOCKER_TAG=2.26.0
+  ${HZN_AIC} start  ${HORIZON_NUM} /tmp/horizon
+  sleep 10
+  docker cp /tmp/exports_list horizon$HORIZON_NUM:/tmp/exports_list
+  docker cp /root/. horizon$HORIZON_NUM:/root/.
```
dabooz, line 34: What are you really copying here?

dabooz (changes requested):

test/gov/init_exchange.sh
dabooz: Defer to future issue

test/gov/verify_agreements.sh
dabooz: Defer to future issue, this logic will be re-written.

test/gov/run_agent.sh
dabooz: Defer to future issue when the code is changed to hzn register, much of this will change anyway.

test/Makefile
njabbouribm: test

linggao: yes, ess related files like auth and .sock file are put under this directory so that the services can access them.

PR #2115, opened 2020-09-28
Fix the reference link error

xauthulei:
The reference link of the `README` can't be accessed, because the files have been moved from the `doc` to `docs`


linggao (approved):

xauthulei:
@linggao , Thanks for your efforts here.

Issue #2116, opened 2020-09-28
Agbot sometimes using wrong messaging key for cluster nodes 


MaxMcAdam:
The cluster appears to be changing its key immediately after patching the key to the exchange. This prevents the agent from understanding proposals made by the agbot

PR #2119, opened 2020-09-28
Issue #2116 Agbot sometimes using wrong messaging key for cluster nodes

MaxMcAdam:

### DIFF
--- a/exchange/messaging.go
+++ b/exchange/messaging.go
@@ -17,6 +17,7 @@ import (
 	"io/ioutil"
 	"os"
 	"path"
+	"sync"
 )
 
 // This module is used to construct a message that can be sent over an insecure transport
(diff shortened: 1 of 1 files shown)


MaxMcAdam:
Bug explanation: The registration thread starting at the api calls GetKeys(), the key file don’t exist yet, so the registration thread starts generating them. While this is happening, the msg worker happens to check messages. It also calls GetKeys() and sees the file doesn’t exist yet. It starts generating keys also. The registration thread writes its keys to file and patches the pub key to the exchange. Then the msg worker finishes generating keys and overwrites the one generated by the first thread.

exchange/messaging.go
```diff
@@ -385,7 +386,14 @@ func HasKeys() bool {
 var privFileName = "privateMessagingKey.pem"
 var pubFileName = "publicMessagingKey.pem"
 
+var KeyLock *sync.Mutex
```
dabooz, line 389: Both threads could call GetKeys at the same time and both find KeyLock to be nil, then both allocate a new lock and use it, one overwrites the other and you end up with the racing keys problem anyway.

dabooz (changes requested):

exchange/messaging.go
MaxMcAdam: fixed

dabooz (approved):

Issue #2138, opened 2020-10-02
Property name cannot start with  'in' in a constraint


linggao:
For a node policy, like the following, failed to get parsed in anax.
````
{
  "properties": [
    {
      "name": "openhorizon.example",
      "value": "helloworld"
    }
  ],
    "constraints": [
	"intValue >= 3"
    ]
}
````

````
 hzn policy update -f np_test.json
Error: bad HTTP code 500 from POST http://localhost:8510/node/policy: Unable to sync the local db with the exchange node policy. Node policy does not validate. constraint language [intValue >= 3] is not supported.  Error finding an expression in intValue >= 3. Error was: Next expression not found: in >= 3
````
In fact anything thing starts with `in` in a constraint will have the above error.

PR #2141, opened 2020-10-05
Issue #2138 - Property name cannot start with 'in' in a constraint

MaxMcAdam:


MaxMcAdam:
The property cannot just be "in" but it can start with "in" now

linggao (dismissed):

linggao (approved):

PR #2142, opened 2020-10-05
ORG_ID

njabbouribm:
Signed-off-by: Nadim Jabbour <Nadim.Jabbour@ibm.com>

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,6 +3,7 @@ module github.com/open-horizon/anax
 go 1.14
 
 require (
+	4d63.com/gochecknoglobals v0.0.0-20200924065721-6443f1912fff // indirect
 	4d63.com/gochecknoinits v0.0.0-20200108094044-eb73b47b9fc4 // indirect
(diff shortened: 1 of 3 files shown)


cgiroua:
Looks good, can remove the 'HERE' debug statement

FYI for others this is a temporary bridge we need (as agbots have primarily been named based off the default org) ... we're now just calling them 'agbot' out of the box.

Passing in ORG_ID will become the new norm, but we still have build artifacts that will leverage the 'ORG_ID-agbot' setup while we transition.

njabbouribm:
> Looks good, can remove the 'HERE' debug statement
> 
> FYI for others this is a temporary bridge we need (as agbots have primarily been named based off the default org) ... we're now just calling them 'agbot' out of the box.
> 
> Passing in ORG_ID will become the new norm, but we still have build artifacts that will leverage the 'ORG_ID-agbot' setup while we transition.

Good catch, one of them must have slipped through the cracks.

linggao (approved):

PR #2144, opened 2020-10-06
Issue 2097 - Anax breaks docker logging

sirotsinskuy:

### DIFF
--- a/cli/service/service.go
+++ b/cli/service/service.go
@@ -6,6 +6,7 @@ import (
 	"github.com/open-horizon/anax/api"
 	"github.com/open-horizon/anax/apicommon"
 	"github.com/open-horizon/anax/cli/cliutils"
+	"github.com/open-horizon/anax/containermessage"
 	"github.com/open-horizon/anax/cutil"
 	"github.com/open-horizon/anax/exchange"
 	"github.com/open-horizon/anax/i18n"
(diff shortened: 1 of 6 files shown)


container/container.go
```diff
@@ -261,10 +261,16 @@ func (w *ContainerWorker) finalizeDeployment(agreementId string, deployment *con
...
+			logDriver = service.LogDriver
+		}
+
 		if !deployment.ServicePattern.IsShared("singleton", serviceName) {
 			labels[LABEL_PREFIX+".agreement_id"] = agreementId
 			logConfig = docker.LogConfig{
-				Type: "syslog",
+				Type: logDriver,
 				Config: map[string]string{
 					"tag": fmt.Sprintf("workload-%v_%v", strings.ToLower(agreementId), serviceName),
```
MaxMcAdam, line 275: I got this error when the log_driver is set to 'local' `API error (400): unknown log opt 'tag' for log driver local.`

cli/service/service.go
```diff
@@ -93,42 +94,69 @@ func Log(serviceName string, tailing bool) {
...
+		if (def.SpecRef == name && def.Org == org) || strings.Contains(def.SpecRef, refUrl) {
+			if def.Deployment != "" {
+				deployment := &containermessage.DeploymentDescription{}
+				if err := json.Unmarshal([]byte(def.Deployment), deployment); err != nil {
+					cliutils.Fatal(cliutils.CLI_INPUT_ERROR, msgPrinter.Sprintf("Deployment unmarshalling error: %v", err))
+				}
+
+				for _, service := range deployment.Services {
+					if service.LogDriver != "" && service.LogDriver != "syslog" {
+						cliutils.Verbose(msgPrinter.Sprintf("Non default log-driver used for service: %v", service.LogDriver))
```
MaxMcAdam, line 145: Change to something like "Service logs are only available for services with 'syslog' log-driver. This service is using log-driver %v".
That way they know which driver supports logs and there's no chance they will think default refers to the docker default not ours.

MaxMcAdam (changes requested):

cli/service/service.go
sirotsinskuy: Changed message, thanks

container/container.go
sirotsinskuy: Yeah, as it turns out, some logging drivers doesn't support tags, and some of them don't like unknown options (json-file for example is just ignoring them). I've added removing of "tag" config for unsupported drivers (this looks simpler than adding few new new `if` blocks for preventing adding of "tag")

Issue #2149, opened 2020-10-07
agent-install.sh update other edge cluster resources when agent image version is unchanged


bmpotter:
The ability to change other resources and make the decision about whether it needs to re-register is based on the `update_cluster()` flow. But `update_cluster()` is too destructive (always deleting/recreating every resource that exists, including the horizon agent deployment).

- [x] Change the `update_cluster()` flow to not delete/recreate the agent deployment if it is already running the version they want, and if it doesn't need to be deleted/recreated to pick up changes in other resources.

PR #2153, opened 2020-10-07
Issue #2149 - Agent install update

LiilyZhang:
Signed-off-by: zhangl  <zhangl@us.ibm.com>

### DIFF
--- a/agent-install/agent-install.sh
+++ b/agent-install/agent-install.sh
@@ -824,6 +824,7 @@ function store_cert_file_permanently() {
 # For both device and cluster: Returns true (0) if /etc/default/horizon already has these values
 function is_horizon_defaults_correct() {
     log_debug "is_horizon_defaults_correct() begin"
+    IS_HORIZON_DEFAULTS_CORRECT='false'
     local anax_port=$1   # optional
     local cert_file defaults_file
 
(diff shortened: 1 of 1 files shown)


agent-install/agent-install.sh
```diff
@@ -824,6 +824,7 @@ function store_cert_file_permanently() {
 # For both device and cluster: Returns true (0) if /etc/default/horizon already has these values
 function is_horizon_defaults_correct() {
     log_debug "is_horizon_defaults_correct() begin"
+    IS_HORIZON_DEFAULTS_CORRECT=1
```
bmpotter, line 827: I'd like you to change this value to `'true'` or `'false'` . When i look at this i don't know if `1` means it is correct or is not correct.

agent-install/agent-install.sh
```diff
@@ -2196,45 +2196,48 @@ function create_deployment() {
...
-        else
-            AGENT_POD_STATUS=$($KUBECTL -n $AGENT_NAMESPACE get pod $POD_ID 2>/dev/null | grep -E '^agent-' | cut -d " " -f3)
-            if [[ $AGENT_POD_STATUS == "Terminating" ]]; then
-                if [[ $AGENT_SKIP_PROMPT == 'false' ]]; then
-                    echo "Agent pod ${POD_ID} is still in Terminating, force delete pod?[y/N]:"
-                    read RESPONSE
-                    if [[ "$RESPONSE" == 'y' ]]; then
-                        log_verbose "Force deleting agent pod ${POD_ID}..."
+    if [[ "$IS_AGENT_IMAGE_VERSION_SAME" == "true" ]] && [[ "$IS_HORIZON_ORG_ID_SAME" == "true" ]]; then
+        log_verbose "Agent image version and value of HZN_ORG_ID are same, will skip updating agent deployment"
```
bmpotter, line 2200: I think this is important enough information that it should be `log_info` instead of `log_verbose`. I would prefer the statement to be `log_info "Agent image version and HZN_ORG_ID are the same. Keeping the existing agent deployment."`

bmpotter (changes requested):

bmpotter (approved):

Issue #2157, opened 2020-10-08
Agbot not using cached node policies


MaxMcAdam:

dlarson04:
verified that is working as expected in build amd64_agbot:2.27.0-142.202010090412.be199123

PR #2158, opened 2020-10-08
Issue #2157 - Agbot not using cached node policies

MaxMcAdam:


MaxMcAdam:
The function to retrieve node policies from the cache asserts the cached object as a .(ExchangePolicy) but the policies were being cached as a pointer making the assertion false so the function was returning nil everytime

linggao (approved):

PR #2178, opened 2020-10-22
Issue 2169 - Node registration with special unicode character

sirotsinskuy:

### DIFF
--- a/cli/register/register.go
+++ b/cli/register/register.go
@@ -72,7 +72,7 @@ func DoIt(org, pattern, nodeIdTok, userPw, inputFile string, nodeOrgFromFlag str
 	msgPrinter := i18n.GetMessagePrinter()
 
 	// check the input
-	org, pattern, waitService, waitOrg = verifyRegisterParamters(org, pattern, nodeOrgFromFlag, patternFromFlag, waitService, waitOrg)
+	org, pattern, waitService, waitOrg = verifyRegisterParamters(org, pattern, nodeOrgFromFlag, patternFromFlag, waitService, waitOrg, nodeIdTok)
 
 	cliutils.SetWhetherUsingApiKey(nodeIdTok) // if we have to use userPw later in NodeCreate(), it will set this appropriately for userPw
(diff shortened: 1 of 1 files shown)


cli/register/register.go
```diff
@@ -629,6 +629,17 @@ func verifyRegisterParamters(org, pattern, nodeOrgFromFlag string, patternFromFl
 		}
 	}
 
+	if nodeId, _ := cliutils.SplitIdToken(nodeIdTok); nodeId != "" {
+		illegalInput, err := api.InputIsIllegal(nodeId)
+		if err != nil {
+			cliutils.Fatal(cliutils.CLI_GENERAL_ERROR, msgPrinter.Sprintf("could not validate node ID '%s': %v", nodeId, err))
+		} else if illegalInput != "" {
+			cliutils.Fatal(cliutils.CLI_INPUT_ERROR, msgPrinter.Sprintf("invalid node ID: A-Za-z0-9 and special symbols !-*+()?.,:&@ are only allowed"))
```
linggao, line 637: "invalid node ID: A-Za-z0-9, unicode characters and special symbols !-*+()?.,:&@ are only allowed"

cli/register/register.go
```diff
@@ -629,6 +629,17 @@ func verifyRegisterParamters(org, pattern, nodeOrgFromFlag string, patternFromFl
 		}
 	}
 
+	if nodeId, _ := cliutils.SplitIdToken(nodeIdTok); nodeId != "" {
```
linggao, line 632: Should also consider getting node id from the agent if the nodeId is empty. See line 129-135. Agent gets the node id from file /etc/default/horizon. User could put a bad id there.

linggao (changes requested):
per comments

cli/register/register.go
sirotsinskuy: Added `nodeId` fetching from agent in case of empty `nodeId`

linggao (approved):

PR #2179, opened 2020-10-26
Issue 2155 - 'hzn unregister -fD' when in unconfiguring state

sirotsinskuy:


api/path_node.go
```diff
@@ -334,7 +334,7 @@ func DeleteHorizonDevice(removeNode string,
 		return errorhandler(NewNotFoundError("The node is not registered.", "node"))
 	} else if !pDevice.IsState(persistence.CONFIGSTATE_CONFIGURED) && !pDevice.IsState(persistence.CONFIGSTATE_CONFIGURING) {
 		LogDeviceEvent(db, persistence.SEVERITY_ERROR, persistence.NewMessageMeta(EL_API_ERR_NODE_UNREG_NOT_IN_STATE), persistence.EC_ERROR_NODE_UNREG, pDevice)
-		return errorhandler(NewBadRequestError(fmt.Sprintf("The node must be in configured or configuring state in order to unconfigure it.")))
+		return errorhandler(NewBadRequestError(fmt.Sprintf("Invalid node state: the node must be in configured or configuring state in order to unconfigure it.")))
```
linggao, line 337: Since the string might get translated, please do not use string comparison. You can make it look like an id so that it will not be translated. Like "INVALIDE_NODE_STATE. The node must be ....".

linggao (changes requested):
per comment

api/path_node.go
sirotsinskuy: Thanks for suggestion. Updated error message

linggao (approved):

PR #2192, opened 2020-11-06
Issue 794 - hzn cli should prevent version mismatches with the Exchange

sirotsinskuy:

### DIFF
--- a/cli/exchange/version.go
+++ b/cli/exchange/version.go
@@ -3,17 +3,31 @@ package exchange
 import (
 	"fmt"
 	"github.com/open-horizon/anax/cli/cliutils"
+	"strings"
 )
 
-func Version(org, userPw string) {
-	cliutils.SetWhetherUsingApiKey(userPw)
-	credToUse := ""
+func Version(org string, userPw string) {
+	output := LoadExchangeVersion(true, org, userPw)
+	fmt.Println(output)
+}
+
+// Loads exchange version based on optional org and optional set of prioritized credentials (first non-empty credentials will be used)
+// If loadWithoutCredentials is false and credentials are missing - the version loading will not be executed (empty string will be returned)
+func LoadExchangeVersion(loadWithoutCredentials bool, org string, userPws ...string) string {
+	var credToUse string
+	for _, cred := range userPws {
+		if cred != "" {
+			credToUse = cred
+			break
+		}
+	}
 	var output []byte
 	// Note: the base exchange does not need creds for this call (although is tolerant of it), but some front-ends to the exchange might
-	if userPw != "" {
-		cliutils.ExchangeGet("Exchange", cliutils.GetExchangeUrl(), "admin/version", cliutils.OrgAndCreds(org, userPw), []int{200}, &output)
-	} else {
+	if credToUse != "" {
+		cliutils.SetWhetherUsingApiKey(credToUse)
+		cliutils.ExchangeGet("Exchange", cliutils.GetExchangeUrl(), "admin/version", cliutils.OrgAndCreds(org, credToUse), []int{200}, &output)
+	} else if loadWithoutCredentials {
 		cliutils.ExchangeGet("Exchange", cliutils.GetExchangeUrl(), "admin/version", credToUse, []int{200}, &output)
 	}
-	fmt.Print(string(output))
+	return strings.TrimSpace(string(output))
 }
--- a/cli/hzn.go
+++ b/cli/hzn.go
@@ -4,6 +4,7 @@ package main
 import (
 	"flag"
 	"github.com/open-horizon/anax/cli/sdo"
+	"github.com/open-horizon/anax/version"
 	"os"
 	"strings"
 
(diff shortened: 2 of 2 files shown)


cli/hzn.go
```diff
@@ -727,6 +728,10 @@ Environment Variables:
 			// get HZN_EXCHANGE_USER_AUTH as default if exUserPw is empty
 			exUserPw = cliutils.RequiredWithDefaultEnvVar(exUserPw, "HZN_EXCHANGE_USER_AUTH", msgPrinter.Sprintf("exchange user authentication must be specified with either the -u flag or HZN_EXCHANGE_USER_AUTH"))
 		}
+
+		if err := version.VerifyExchangeVersion1(exchange.LoadExchangeVersion(*exOrg, *exUserPw), false, true); err != nil {
```
linggao, line 732: Some commands use -n instead of -u for cred. So the code here should use credToUse if it is not empty, otherwise use exUserPw.

cli/hzn.go
```diff
@@ -735,6 +740,10 @@ Environment Variables:
 
 		// use HZN_EXCHANGE_NODE_AUTH for -n and trim the org
 		nodeIdTok = cliutils.WithDefaultEnvVar(nodeIdTok, "HZN_EXCHANGE_NODE_AUTH")
+
+		if err := version.VerifyExchangeVersion1(exchange.LoadExchangeVersion(*org, *userPw), false, true); err != nil {
```
linggao, line 744: Here is opposite, -n is usually specified the node name and token to be created with -u cred. But it could also be used as a cred if the node already exists. So the logic should be: use userPw if it is not empty, else use nodeIdTok if it is not empty. If both are empty, just skip the version check.

cli/hzn.go
```diff
@@ -758,6 +767,10 @@ Environment Variables:
 		if *allCompBPolFile == "" {
 			allCompBPolFile = allCompDepPolFile
 		}
+
+		if err := version.VerifyExchangeVersion1(exchange.LoadExchangeVersion(*deploycheckOrg, *deploycheckUserPw), false, true); err != nil {
```
linggao, line 771: For `hzn deploycheck` there is a case when deploycheckUserPw is not required when all the inputs are local files. So  please skip the version check if deploycheckUserPw is empty.

version/version.go
```diff
@@ -39,7 +39,11 @@ func VerifyExchangeVersion1(exch_version string, checkWithPreferred bool) error
 	} else if comp, err := semanticversion.CompareVersions(exch_version, version_for_check); err != nil {
 		return fmt.Errorf("Failed to compare the versions. %v", err)
 	} else if comp < 0 {
-		return fmt.Errorf("The current exchange version %v does not meet the requirement. The required version is %v or above. Please upgrade the exchange.", exch_version, version_for_check)
+		return fmt.Errorf("The current exchange version %v does not meet the requirement. The required version is %v or above. "+
+			"Please upgrade the exchange.", exch_version, version_for_check)
+	} else if checkExactVersion && comp > 0 {
```
linggao, line 44: This parts needs more discussion. Please start a slack group discussion with @dabooz, @TheMosquito and myself.

cli/hzn.go
sirotsinskuy: Implementation has been changed. Please look into, does it look OK now?

cli/hzn.go
```diff
@@ -735,6 +742,12 @@ Environment Variables:
 
 		// use HZN_EXCHANGE_NODE_AUTH for -n and trim the org
 		nodeIdTok = cliutils.WithDefaultEnvVar(nodeIdTok, "HZN_EXCHANGE_NODE_AUTH")
+
```
linggao, line 745: Please add `org = cliutils.WithDefaultEnvVar(org, "HZN_ORG_ID")`  here so that the org can get the HZN_ORG_ID env.

cli/exchange/version.go
```diff
@@ -3,17 +3,31 @@ package exchange
...
+	fmt.Println(output)
+}
+
+// Loads exchange version based on optional org and optional set of prioritized credentials (first non-empty credentials will be used)
+// If loadWithoutCredentials is false and credentials are missing - the version loading will not be executed (empty string will be returned)
+func LoadExchangeVersion(loadWithoutCredentials bool, org string, userPws ...string) string {
+	var credToUse string
+	for _, cred := range userPws {
+		if cred != "" {
+			credToUse = cred
```
linggao, line 20: Need a `break` once cred is not a empty string.

cli/exchange/version.go
```diff
@@ -3,17 +3,31 @@ package exchange
...
+		}
+	}
+
+	cliutils.SetWhetherUsingApiKey(credToUse)
```
linggao, line 24: credToUse could be empty. Do we need this?

linggao (changes requested):
per comments

cli/exchange/version.go
sirotsinskuy: Moved to non-empty creds section

cli/exchange/version.go
sirotsinskuy: Added a break, thanks

cli/hzn.go
sirotsinskuy: Added new local org variable for version check, because assigning some value to `org` from env var is producing errors - we cannot have it set for some cases (`-o and -p are mutually exclusive with <nodeorg> and <pattern> arguments`)

linggao (approved):

PR #2203, opened 2020-11-17
Issue 2095 - Bidirectional communication with dependencies

dabooz:
Signed-off-by: Dave Booz <booz@us.ibm.com>

### DIFF
--- a/container/container.go
+++ b/container/container.go
@@ -751,7 +751,7 @@ func serviceStart(client *docker.Client,
 	}
 	if serviceConfig.HostConfig.NetworkMode != "host" {
 		for _, cfg := range sharedEndpoints {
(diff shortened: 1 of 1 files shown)


container/container.go
```diff
@@ -1106,7 +1106,7 @@ func (b *ContainerWorker) ResourcesCreate(agreementId string, agreementProtocol
 			}
 
 			woutAliases := &docker.EndpointConfig{
```
sirotsinskuy, line 1108: It would be good to rename `woutAliases` to something else, since there are aliases now

dabooz: I agree

dabooz:
This was temporary just to communicate a possible fix.

PR #2204, opened 2020-11-20
Fix "hzn dev service new" command to publish services with "public":false

clementkng:
This commit partially addresses this [open-horizon/examples issue](https://github.com/open-horizon/examples/issues/356). Services should be created with `"public": false` so those outside the organizations can't see them by default.

Tested by locally building the hzn cli and making sure that the resulting horizon files have `"public": false` set on them.

Signed-off-by: Clement Ng <clementdng@gmail.com>

### DIFF
--- a/cli/dev/servicedefinition.go
+++ b/cli/dev/servicedefinition.go
@@ -44,7 +44,7 @@ func CreateServiceDefinition(directory string, specRef string, imageInfo map[str
 
 	res.Org = "$HZN_ORG_ID"
 	res.Label = msgPrinter.Sprintf("$SERVICE_NAME for $ARCH")
-	res.Public = true
+	res.Public = false
 	res.URL = "$SERVICE_NAME"
 	res.Version = "$SERVICE_VERSION"
 	res.Arch = "$ARCH"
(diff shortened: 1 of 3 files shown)


dabooz:
@clementdng please open an issue in the anax repo for this change, connect this PR to that issue, then attached the PR submitted label and move the issue into review/QA state.

dabooz (approved):

clementkng:
@dabooz thanks for reviewing this. Sorry, I think there's still some discussion in the [open-horizon/examples issue](https://github.com/open-horizon/examples/issues/356) so I'm not sure whether this is the exact behavior we want. I may convert this to a draft so it doesn't get accidentally merged in.

linggao (changes requested):
This change broke the e2edev test. Please fix test/gov/hzn_dev_services.sh to make the service public.

linggao:
@clementkng Please see my comment above. Need to fix test/gov/hzn_dev_services.sh to make the services public.

clementkng:
@linggao thanks for reviewing! Right now, I'm working on understanding the e2edev test so I can make the test `Helloservice` service public. In the meantime, it seems that the original issue in the examples repo has changed direction, but I still think this PR is still helpful for Open Horizon users.

linggao:
@bmpotter do we still need this PR?

bmpotter (approved):

bmpotter:
> @bmpotter do we still need this PR?

@linggao yes we do

clementkng:
@linggao I was able to fix the `test/gov/hzn_dev_services.sh`

linggao (approved):

PR #2207, opened 2020-11-25
Issue 2211 - Add public flag to hzn exchange service publish

clementkng:
This commit addresses #2211. This adds a flag to the `hzn exchange service publish` that sets the `public` field of the service to either be true or false based on the chart below. The rules are as follows.

1.  If the `hzn exchange service publish` has the flag set, then it takes precedence
2. Otherwise, `public` defaults to whatever is set in the metadata ie `horizon/service.definition.json`
3. If both are undefined, set to `public` to false

image

Tested by locally building hzn cli and making sure that running `hzn exchange service publish` follows the results above.

Signed-off-by: Clement Ng <clementdng@gmail.com>

### DIFF
--- a/cli/exchange/service.go
+++ b/cli/exchange/service.go
@@ -120,7 +120,7 @@ func ServiceList(credOrg, userPw, service string, namesOnly bool, filePath strin
 }
 
 // ServicePublish signs the MS def and puts it in the exchange
-func ServicePublish(org, userPw, jsonFilePath, keyFilePath, pubKeyFilePath string, dontTouchImage bool, pullImage bool, registryTokens []string, overwrite bool, servicePolicyFilePath string) {
+func ServicePublish(org, userPw, jsonFilePath, keyFilePath, pubKeyFilePath string, dontTouchImage bool, pullImage bool, registryTokens []string, overwrite bool, servicePolicyFilePath string, public string) {
 	// get message printer
 	msgPrinter := i18n.GetMessagePrinter()
 
@@ -138,6 +138,19 @@ func ServicePublish(org, userPw, jsonFilePath, keyFilePath, pubKeyFilePath strin
 	}
 	if svcFile.Org != "" && svcFile.Org != org {
 		cliutils.Fatal(cliutils.CLI_INPUT_ERROR, msgPrinter.Sprintf("the org specified in the input file (%s) must match the org specified on the command line (%s)", svcFile.Org, org))
+	}
+	if public != "" {
+		public = strings.ToLower(public)
+		if public != "true" && public != "false" {
+			cliutils.Fatal(cliutils.CLI_INPUT_ERROR, msgPrinter.Sprintf("Need to set 'true' or 'false' when specifying flag --public."))
+		} else {
+			// Override public key with key set in the hzn command
+			svcFile.Public, err = strconv.ParseBool(public)
+			if err != nil {
+				cliutils.Fatal(cliutils.INTERNAL_ERROR, msgPrinter.Sprintf("failed to parse %s: %v", public, err))
+			}
+
+		}
 	}
 
 	// Compensate for old service definition files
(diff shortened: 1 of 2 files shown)


dabooz:
@clementkng please open an issue for this change, connect this PR to the issue, set the PR submitted label and move the issue into Review/QA state. thanks.

clementkng:
@dabooz I don't think I have permission to move the issue into Review/QA state, but I did do the rest of the steps.
image

t-fine:
@clementkng oh of course let me get you introduced to the zen board! I'll show you in our meeting today

cli/hzn.go
```diff
@@ -281,6 +281,7 @@ Environment Variables:
 	exSvcRegistryTokens := exServicePublishCmd.Flag("registry-token", msgPrinter.Sprintf("Docker registry domain and auth that should be stored with the service, to enable the Horizon edge node to access the service's docker images. This flag can be repeated, and each flag should be in the format: registry:user:token")).Short('r').Strings()
 	exSvcOverwrite := exServicePublishCmd.Flag("overwrite", msgPrinter.Sprintf("Overwrite the existing version if the service exists in the Exchange. It will skip the 'do you want to overwrite' prompt.")).Short('O').Bool()
 	exSvcPolicyFile := exServicePublishCmd.Flag("service-policy-file", msgPrinter.Sprintf("The path of the service policy JSON file to be used for the service to be published. This flag is optional")).Short('p').String()
+	exSvcPublic := exServicePublishCmd.Flag("public", msgPrinter.Sprintf("Whether the service is visible to users outside of the organization. This flag is optional. If left unset, the service will default to whatever the metadata has set. If the metadata has also not set the public key, then the service will by default not be public.")).Short('V').String()
```
dabooz, line 284: minor nit; the term "public key" means something different in the context of publishing a service, so to avoid potential confusion, please update the last sentence to "If the service definition has also not set the public field, then the service will by default not be public."

dabooz (changes requested):

cli/exchange/service.go
```diff
@@ -139,6 +139,15 @@ func ServicePublish(org, userPw, jsonFilePath, keyFilePath, pubKeyFilePath strin
 	if svcFile.Org != "" && svcFile.Org != org {
 		cliutils.Fatal(cliutils.CLI_INPUT_ERROR, msgPrinter.Sprintf("the org specified in the input file (%s) must match the org specified on the command line (%s)", svcFile.Org, org))
 	}
+	if public != "" {
+		public = strings.ToLower(public)
+		if public != "true" && public != "false" {
+			cliutils.Fatal(cliutils.CLI_INPUT_ERROR, msgPrinter.Sprintf("Need to set 'true' or 'false' when specifying flag -V."))
+		} else {
+			// Override public key with key set in the hzn command
+			svcFile.Public, _ = strconv.ParseBool(public)
```
dabooz, line 148: I didnt notice this last review. If you ignore the error on the string convert, and there is an error, the pubic field will get set to false. I think you need to check for an error here and report it as fatal.

cli/hzn.go
```diff
@@ -281,6 +281,7 @@ Environment Variables:
 	exSvcRegistryTokens := exServicePublishCmd.Flag("registry-token", msgPrinter.Sprintf("Docker registry domain and auth that should be stored with the service, to enable the Horizon edge node to access the service's docker images. This flag can be repeated, and each flag should be in the format: registry:user:token")).Short('r').Strings()
 	exSvcOverwrite := exServicePublishCmd.Flag("overwrite", msgPrinter.Sprintf("Overwrite the existing version if the service exists in the Exchange. It will skip the 'do you want to overwrite' prompt.")).Short('O').Bool()
 	exSvcPolicyFile := exServicePublishCmd.Flag("service-policy-file", msgPrinter.Sprintf("The path of the service policy JSON file to be used for the service to be published. This flag is optional")).Short('p').String()
+	exSvcPublic := exServicePublishCmd.Flag("public", msgPrinter.Sprintf("Whether the service is visible to users outside of the organization. This flag is optional. If left unset, the service will default to whatever the metadata has set. If the service definition has also not set the public field, then the service will by default not be public.")).Short('V').String()
```
dabooz, line 284: I dont like using -V, I would prefer that there was no short flag so the user has to specify --public.

dabooz (changes requested):

dabooz (approved):

PR #2214, opened 2020-12-07
Issue 2188 - docker login credentials in the OS keyring

sirotsinskuy:

### DIFF
--- a/cli/cliutils/cliutils.go
+++ b/cli/cliutils/cliutils.go
@@ -204,7 +204,8 @@ func NewDockerClient() (client *dockerclient.Client) {
 	return
 }
 
-// GetDockerAuth finds the docker credentials for this registry in ~/.docker/config.json
+// GetDockerAuth finds the docker credentials for this registry in ~/.docker/config.json.
+// It also will try to obtains credentials from a docker credential store if it's in use.
 func GetDockerAuth(domain string) (auth dockerclient.AuthConfiguration, err error) {
 	var auths *dockerclient.AuthConfigurations
 	if auths, err = dockerclient.NewAuthConfigurationsFromDockerCfg(); err != nil {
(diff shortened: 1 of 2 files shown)


cli/cliutils/cliutils.go
```diff
@@ -214,6 +215,49 @@ func GetDockerAuth(domain string) (auth dockerclient.AuthConfiguration, err erro
 		}
 	}
 
+	// try to load cred from configured credential store
```
dabooz, line 218: Please put a comment on the GetDockerAuth function indicating that it also obtains creds from a docker credential store if it's in use.

cli/cliutils/cliutils.go
```diff
@@ -214,6 +215,49 @@ func GetDockerAuth(domain string) (auth dockerclient.AuthConfiguration, err erro
...
+	var cfgPath string
+	if dockerCfgPath := os.Getenv("DOCKER_CONFIG"); dockerCfgPath != "" {
```
dabooz, line 220: Why didnt you use the docker auth from cred store function in go-dockerclient? "https://github.com/fsouza/go-dockerclient/blob/main/auth.go#L296"

dabooz (changes requested):

cli/cliutils/cliutils.go
sirotsinskuy: In the current version of go-dockerclient we're using this functionality isn't available. Can I update version to the latest - 1.7.2?

dabooz: Yes, I think that would be a more appropriate way to address the problem.

dabooz (changes requested):
Ok, this looks like it solves the problem for hzn exchange service publish when it has to manipulate the image, but what about the agent at runtime? Does the CLI code push the creds from the docker cred store to the exchange when the service is published? If so, I am not seeing it in the code.

dabooz (dismissed):

linggao (approved):

Issue #2187, opened 2020-11-03
'edgeNodeFiles.sh -f <dir>' does not work


bmpotter:
It puts the gathered files in the current directory instead of the one specified.

The fix is that all 3 lines:

```
if [[ -n $DIR ]]; then pushd; fi
```

need to be changed to this:

```
if [[ -n $DIR ]]; then pushd $DIR; fi
```

Until this is fixed a work around is to run it like this:

```
(cd <dir> && edgeNodeFiles.sh ... )
```

xauthulei:
@bmpotter , Is that OK to let me fix this ? Thanks

bmpotter:
@xauthulei Yes, thank you!

PR #2216, opened 2020-12-08
Fix the issue2187

xauthulei:
I am try to address the issue #2187,  @bmpotter , Could you review this, thanks

### DIFF
--- a/agent-install/edgeNodeFiles.sh
+++ b/agent-install/edgeNodeFiles.sh
@@ -511,7 +511,7 @@ all_main() {
 
     cloudLogin
 
-    if [[ -n $DIR ]]; then pushd; fi   # if they want the files somewhere else, make that our current dir
+    if [[ -n $DIR && -d $DIR ]]; then pushd $DIR; fi   # if they want the files somewhere else, make that our current dir
 
     cleanUpPreviousFiles
 
@@ -534,7 +534,7 @@ all_main() {
         createTarFile
     fi
 
-    if [[ -n $DIR ]]; then popd; fi
+    if [[ -n $DIR && -d $DIR ]]; then popd; fi
 }
 
 cluster_main() {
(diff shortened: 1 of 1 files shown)


agent-install/edgeNodeFiles.sh
```diff
@@ -511,7 +511,7 @@ all_main() {
 
     cloudLogin
 
-    if [[ -n $DIR ]]; then pushd; fi   # if they want the files somewhere else, make that our current dir
+    if [[ -n $DIR && -d $DIR ]]; then pushd $DIR; fi   # if they want the files somewhere else, make that our current dir
```
bmpotter, line 514: This isn't quite right. The effect of `&& -d $DIR` is that if the user specified `-f <dir>` flag (which sets `DIR`) but the `<dir>` specified didn't exist or was a file, it would silently not use it (confusing the user).

You have 2 choices:

1. Just remove `&& -d $DIR` from all of the lines you added it to. This will cause `pushd` to display an error msg if the directory is not valid.
2. Or do number 1 and add a check in `checkPrereqsAndInput` that the directory specified is valid, and display an error msg and return exit code 1 if not.

The fact that you didn't catch this problem makes me think you didn't test your code thoroughly enough before submitting the PR. Please test both the good path and the error cases. Thx.

bmpotter (changes requested):

agent-install/edgeNodeFiles.sh
xauthulei: Yea, it should be reasonable. Thanks.

xauthulei:
@bmpotter , Because my local repo is deleted unintentionally, I can't commit my changes in this PR. I created a new PR https://github.com/open-horizon/anax/pull/2234. Sorry for this inconvenience。

Issue #2191, opened 2020-11-05
Support hubadmin for 'hzn exchange user create' command


linggao:
hzn exchange user create --hubadmin
hzn exchange user sethubadmin
The  underlying  exchange API works this way: For making an existing IAM user a hubAdmin the root user (or another hubAdmin) has to curl the exchange like this:
````
curl -X POST -sS -u "root/root:$EXCHANGE_ROOTPW" "http://$EXCHANGE/orgs/root/users/<IAM-user-id>" -H "accept: application/json" -H "Content-Type: application/json" -d '{"password": "", "admin": false, "hubAdmin": true, "email": ""}' | jq
````
This is basically making a "base" user in the root org for the IAM user with hubAdmin set to true. The catch is that you have to make sure the userid of this user matches the IAM user id. So if in IAM my user is user1 the curl should look like this:
````
curl -X POST -sS -u "root/root:$EXCHANGE_ROOTPW" "http://$EXCHANGE/orgs/root/users/user1" -H "accept: application/json" -H "Content-Type: application/json" -d '{"password": "", "admin": false, "hubAdmin": true, "email": ""}' | jq
````
Otherwise it's just the same as making a normal exchange local user,  you just have to make them in the root org and set hubAdmin to true

PR #2218, opened 2020-12-09
Issue #2191 - Support hubadmin for 'hzn exchange user create' command

MaxMcAdam:


cli/exchange/user.go
```diff
@@ -76,6 +94,35 @@ func UserSetAdmin(org, userPwCreds, user string, isAdmin bool) {
...
 
+func UserSetHubAdmin(org, userPwCreds, user string, isHubAdmin bool) {
+	// get message printer
+	msgPrinter := i18n.GetMessagePrinter()
+
+	_, userId := cliutils.TrimOrg(org, user)
+
+	// Get the user from the exchange
+	var userList ExchangeUsers
+	httpCode := cliutils.ExchangeGet("Exchange", cliutils.GetExchangeUrl(), "orgs/root/users/"+userId, cliutils.OrgAndCreds(org, userPwCreds), []int{200, 404}, &userList)
```
dabooz, line 105: For the case where the exchange API returns an authorization problem, does the user get a reasonable message. I think this case might be somewhat special because the caller MUST be the exchange root user or a hubadmin user, so there should be a message to this effect to help the user.

cli/exchange/user.go
```diff
@@ -76,6 +94,35 @@ func UserSetAdmin(org, userPwCreds, user string, isAdmin bool) {
...
+		} else {
+			updatedUser := cliutils.UserExchangeReq{Password: existingUser.Password, Admin: existingUser.Admin, HubAdmin: isHubAdmin, Email: existingUser.Email}
+			cliutils.ExchangePutPost("Exchange", http.MethodPut, cliutils.GetExchangeUrl(), "orgs/root/users/"+userId, cliutils.OrgAndCreds(org, userPwCreds), []int{201}, updatedUser, nil)
+		}
+	} else if isHubAdmin {
+		// If the user does not exist in the root org and the user isHubAdmin, create them with hubAdmin set to true and no other fields
+		postUserReq := cliutils.UserExchangeReq{HubAdmin: true}
+		httpCode := cliutils.ExchangePutPost("Exchange", http.MethodPost, cliutils.GetExchangeUrl(), "orgs/root/users/"+userId, cliutils.OrgAndCreds(org, userPwCreds), []int{201, 400}, postUserReq, nil)
+		if httpCode == 400 {
+			cliutils.Fatal(cliutils.CLI_GENERAL_ERROR, msgPrinter.Sprintf("Only exchange users in the root org can be hub admins."))
```
dabooz, line 120: I dont understand this error message. We get here because of a 404 on the GET, but we received a 400 because why? we already forced the user into root org based on how the API is being called.

dabooz (changes requested):

cli/exchange/user.go
MaxMcAdam: Changed to explain the issue when a 403 is returned

cli/exchange/user.go
MaxMcAdam: The 400 is what we get if we try to call this post on a user who does not exist in the root org, and is not an IAM user. This post with the email and some other fields blank is only allowed for IAM users. This isn't really clear from the error message here, but I am trying to avoid saying "IAM" in the open source

dabooz (approved):

PR #2234, opened 2021-01-11
Fix the issue2187

xauthulei:


xauthulei:
I am try to address the issue #2187, @bmpotter , Could you review this here, thanks

xauthulei:
Review comments from @bmpotter in [PR2216 ](https://github.com/open-horizon/anax/pull/2216)：


This isn't quite right. The effect of && -d $DIR is that if the user specified -f <dir> flag (which sets DIR) but the <dir> specified didn't exist or was a file, it would silently not use it (confusing the user).

You have 2 choices:

1. Just remove && -d $DIR from all of the lines you added it to. This will cause pushd to display an error msg if the directory is not valid.
   
2. Or do number 1 and add a check in checkPrereqsAndInput that the directory specified is valid, and display an error msg and return exit code 1 if not.

The fact that you didn't catch this problem makes me think you didn't test your code thoroughly enough before submitting the PR. Please test both the good path and the error cases. Thx.

agent-install/edgeNodeFiles.sh
```diff
@@ -150,6 +150,14 @@ function checkPrereqsAndInput () {
...
     echo " - CLUSTER_PW set"
     echo ""
+
+    echo "Checking the value of parament option..."
+    if [[ -d $DIR ]]; then
+	echo ' - Value of parament options valid'
+    else
+	fatal 1 "ERROR: the value of parament option '-f' isn't an existing directory ..."
+    fi 
+    echo ""
```
bmpotter, line 160: Did you test this in the case when `-f <dir>` was **not** specified? I think your code would exit with the error message, when it shouldn't. I already told you that you can't submit PRs without thoroughly testing them first.

1. So your check should be `if [[ -n $DIR && ! -d $DIR ]]; then ...`
2. The check should be in the code section that starts with `echo "Checking command arguments ...", because it we don't need a 2nd good output after " - Command arguments valid", and in the error case it would be confusing to say the command arguments are valid and then give an error message about `-f`
3. The word `parament` is misspelled, but also unnecessary

bmpotter (changes requested):

agent-install/edgeNodeFiles.sh
xauthulei: Yea, it is my fault. I don't test this included that condition,  as it always check the some tools , such as `oc`, `cloudctl`. 

This time, I simply disable those checking, and do following testing:

1. without `-f` option:
```
% ./edgeNodeFiles.sh x86_64-macOS 
Checking command arguments ...
 - Command arguments valid
```
2. `-f` with non-directory:
```
% ls -ld tmp                            
ls: tmp: No such file or directory
 % ./edgeNodeFiles.sh x86_64-macOS -f tmp
Checking command arguments ...
ERROR: the value of option '-f' isn't an existing directory ...
```
3. `-f` with correct directory:
```
% ls -ld horizon-agent-macos-pkg-x86_64 
drwxr-xr-x@ 4 llhu  staff  128 Jan 11 13:22 horizon-agent-macos-pkg-x86_64
% ./edgeNodeFiles.sh x86_64-macOS -f horizon-agent-macos-pkg-x86_64
Checking command arguments ...
 - Command arguments valid
 ```
@bmpotter ,Could you review this again, I will be careful later. Thanks

bmpotter: looks good now, thx

bmpotter (approved):

PR #2241, opened 2021-01-20
Issue 2235 - Agbot failed to give new proposals due to old node cache

linggao:


agreementbot/node_search.go
```diff
@@ -121,6 +123,10 @@ func (n *NodeSearch) Scan() {
 		}
 	}
 
+	// allow clearing the cache for all the exchange resources, the findAndMakeAgreements
+	// function will clear the cache and set it false after it finds devices to make agreements.
+	n.clearExchangeCache = true
```
dabooz, line 128: I think there is a problem setting the flag here. The Scan function can get called while a scan is in progress. If that happens, you will turn the flag on and invalidate during the scan. Also, you have a thread problem here. The scans actually occur on a subthread, so you are setting this flag on a different thread than the one that clears it, which could cause a panic. You need to set this flag inside findAndMakeAgreements because its running on the scan sub-thread.

dabooz (changes requested):

agreementbot/node_search.go
linggao: Done. Thanks for the catch.

dabooz (approved):

Issue #2197, opened 2020-11-11
Make Agbot ProtocolTimeoutS and ExchangeMessageTTL be relative to max ORG heartbeat


dlarson04:
Some customers may change the ORG max heartbeat interval if they don't need updates every 2 minutes to say 30 minutes or an hour but the ProtocolTimeoutS and ExchangeMessageTTL parameters in the agbot config are an absolute value... If a customer changes the max heartbeat to 30 minutes, they have to change these values so that these are longer than the max heartbeat.  This requires an agbot restart and a customer may forget about doing it making it hard for an agent to get an agreement 

One solution is to make these values relative to the org level max heartbeat.  That is if the ProtocolTimeoutS is 360 and the  org max heartbeat is 600 seconds, when the agbot creates a message for the agent in this org, it sets the actual timeout to to be 960 seconds (360 +  600).  Likewise for the TTL of the message.   

If implemented like this, the ProtocolTimeoutS and ExchangeMessageTTL would not have to be changed if a customer increases the org level max heartbeat.

MaxMcAdam:
The plan is to not set ProtocolTimeoutS or ExchangeMessageTTL in the config by default for future releases. Instead we will make those timeout decisions based on a factor that scales relative to the max heartbeat. Since an individual node's max heartbeat can exceed the org one, these timeouts will become agreement specific and be stored in the agreements. To make the changes compatible for upgrade, if either of those fields if present in the config for an agbot, they will continue to be used as is.

dlarson04:
@MaxMcAdam   Found another variable AgreementTimeoutS that defaults to 360 seconds... but if the heartbeat times change to 10 minutes or so, this could easily timeout and get an agreement in progress cancelled. 

Ideally, this one would also be based on the individual nodes value like the others

dlarson04:
Verified using agbot v2.28.0-338

PR #2253, opened 2021-01-26
Issue #2197 - Make Agbot ProtocolTimeoutS and ExchangeMessageTTL be r…

MaxMcAdam:
…elative to max ORG heartbeat

### DIFF
--- a/agreement/agreement.go
+++ b/agreement/agreement.go
@@ -612,6 +612,11 @@ func (w *AgreementWorker) syncOnInit() error {
 			}
 			neededBCInstances[bcOrg][bcType][bcName] = true
 
+			// If the agent has been updated from a version 4.2 or ealier, the agreement timeout field will not be set on agreements and must be set for in-flight agreements to be finalized
+			if ag.AgreementTimeout == 0 {
+				w.setAgreementTimeout(ag)
+			}
+
 			// If there is an active agreement that is marked as terminated, then anax was restarted in the middle of
 			// termination processing, and therefore we dont know how far it got. Initiate a cancel again to clean it up.
 			if ag.AgreementTerminatedTime != 0 {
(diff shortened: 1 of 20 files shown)


dabooz (changes requested):

agreement/agreement.go
```diff
@@ -683,8 +683,9 @@ func (w *AgreementWorker) syncOnInit() error {
 					}
 				}
 				glog.V(3).Infof(logString(fmt.Sprintf("added agreement %v to policy agreement counter.", ag.CurrentAgreementId)))
+			} else if ag.AgreementTimeout == 0 {
```
dabooz, line 686: Normally, every agreement will fall into the previous else if at line 665. I think your new code needs to be a standalone if.  Maybe do it at the top of the loop.

agreementbot/agreementbot.go
```diff
@@ -384,6 +384,40 @@ func (w *AgreementBotWorker) Initialize() bool {
 	return true
 }
 
+func (w *AgreementBotWorker) preprocessExistingAgreeements() {
```
dabooz, line 387: Where is this called? And, I thought you were going to do the checks on the agbot side as needed, instead of updating every agreement.

dabooz (changes requested):

agreement/agreement.go
MaxMcAdam: Moved that piece to it's own if

agreementbot/agreementbot.go
MaxMcAdam: Yeah, I think I changed to that partway through writing it and didn't call the function from either spot. It is fixed now

dabooz (approved):

linggao (dismissed):

linggao (approved):

PR #2270, opened 2021-01-29
Issue 2040 (part1) - Agent-install.sh edge cluster fixes/improvements

LiilyZhang:


agent-install/agent-install.sh
```diff
@@ -91,7 +91,7 @@ Additional Edge Device Variables (in environment or config file):
     AGENT_WAIT_MAX_SECONDS: Maximum seconds to wait for the Horizon agent to start or stop. Default: 30
 
 Additional Edge Cluster Variables (in environment or config file):
-    IMAGE_ON_EDGE_CLUSTER_REGISTRY: (required) agent image path (without tag) in edge cluster registry. For OCP use: <registry-host>/<agent-project>/amd64_anax_k8s, for k3s use: <IP-address>:5000/<agent-namespace>/amd64_anax_k8s, for microsk8s: localhost:32000/<agent-namespace>/amd64_anax_k8s
+    IMAGE_ON_EDGE_CLUSTER_REGISTRY: agent image path (without tag) in edge cluster registry. For OCP use: <registry-host>/<agent-project>/amd64_anax_k8s, for k3s use: <IP-address>:5000/<agent-namespace>/amd64_anax_k8s, for microsk8s: localhost:32000/<agent-namespace>/amd64_anax_k8s
```
bmpotter, line 94: `IMAGE_ON_EDGE_CLUSTER_REGISTRY` is now just here to let the user override the value if they don't like what `agent-install.sh` will default it to. I don't think we should confuse them with all of this information, when they probably don't need to know that. So i the text can be:

IMAGE_ON_EDGE_CLUSTER_REGISTRY: override the agent image path (without tag) if you want it to be different from what this script will default it to.

bmpotter (changes requested):

bmpotter (approved):

PR #2279, opened 2021-02-03
Issue 2243 - Add Power (ppc64le) support to agent-install.sh

### LINKED ISSUE #2243: Add Power (ppc64le) support to agent-install.sh
- [x] OS support: RHEL 7.9 (can RHEL 8.x or CentOS be added too?)
- [x] Add edge device support
- [x] Add edge cluster support: the anax k8s dockerfile has been updated to support ppc64le. Are changes needed to the overall horizon support and agent operator to work in a ppc64le kube?
- [x] Need to update `edgeNodeFiles.sh` to put ppc64le pkg and images in CSS and tar file

jacob-rogers:
Signed-off-by: Dmitry Gorbachev <dmitry.gorbachev@ru.ibm.com>

Resolves #2243

### DIFF
--- a/agent-install/agent-install.sh
+++ b/agent-install/agent-install.sh
@@ -26,7 +26,7 @@ SUPPORTED_DEBIAN_ARCH=(amd64 arm64 armhf $SUPPORTED_DEBIAN_ARCH_APPEND)   # comp
 SUPPORTED_REDHAT_VARIANTS=(rhel centos $SUPPORTED_REDHAT_VARIANTS_APPEND)   # compared to what our detect_distro() sets DISTRO to
 # Note: RHEL 8.3 is not officially supported yet, but is enabled only for testing and tech preview purposes
 SUPPORTED_REDHAT_VERSION=(8.1 8.2 8.3 $SUPPORTED_REDHAT_VERSION_APPEND)   # compared to what our detect_distro() sets DISTRO_VERSION_NUM to. For fedora versions see https://fedoraproject.org/wiki/Releases,
-SUPPORTED_REDHAT_ARCH=(x86_64 aarch64 $SUPPORTED_REDHAT_ARCH_APPEND)   # compared to uname -m
+SUPPORTED_REDHAT_ARCH=(x86_64 aarch64 ppc64le $SUPPORTED_REDHAT_ARCH_APPEND)   # compared to uname -m
 
 HOSTNAME=$(hostname -s)
 MAC_PACKAGE_CERT="horizon-cli.crt"
@@ -812,7 +812,7 @@ function download_pkgs_from_anax_release() {
     log_debug "download_pkgs_from_anax_release() begin"
     # This function is called if INPUT_FILE_PATH starts with at least https://github.com/open-horizon/anax/releases
     # Note: adjust_input_file_path() has already been called, which applies some default (if necessary) to INPUT_FILE_PATH
-    local tar_file_name="horizon-agent-${OS}-$(get_pkg_type)-$ARCH.tar.gz"
+    local tar_file_name="horizon-agent-${OS}-$(get_pkg_type)-${ARCH}.tar.gz"
     local remote_path="${INPUT_FILE_PATH%/}/$tar_file_name"
 
     # Download and unpack the package tar file
@@ -833,7 +833,7 @@ function download_pkgs_from_css() {
     log_debug "download_pkgs_from_css() begin"
     # This function is called if INPUT_FILE_PATH starts with css: . We have to add in HZN_FSS_CSSURL to the URL we download from.
     # Note: adjust_input_file_path() has already been called, which applies some default (if necessary) to INPUT_FILE_PATH
-    local tar_file_name="horizon-agent-${OS}-$(get_pkg_type)-$ARCH.tar.gz"
+    local tar_file_name="horizon-agent-${OS}-$(get_pkg_type)-${ARCH}.tar.gz"
 
     # Download and unpack the package tar file
     download_css_file "$INPUT_FILE_PATH/$tar_file_name"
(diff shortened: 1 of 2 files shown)


agent-install/agent-install.sh
```diff
@@ -94,7 +94,11 @@ Additional Edge Device Variables (in environment or config file):
     AGENT_WAIT_MAX_SECONDS: Maximum seconds to wait for the Horizon agent to start or stop. Default: 30
 
 Additional Edge Cluster Variables (in environment or config file):
+<<<<<<< HEAD
```
bmpotter, line 97: This is a merge conflict. Please resolve it by keeping what is in the HEAD and update the PR

agent-install/agent-install.sh
```diff
@@ -1823,7 +1842,7 @@ function loadClusterAgentImage() {
             if [[ -z $image_tag ]]; then
                 image_tag='latest'
             fi
-            local image_path="openhorizon/amd64_anax_k8s:$image_tag"
+            local image_path="openhorizon/${ARCH}_anax_k8s:$image_tag"
```
bmpotter, line 1845: - On a rhel edge cluster ARCH will be `x86_64` instead of `amd64`, right? I don't think we have published `openhorizon/x86_64_anax_k8s:<tag>`
- Did you test this on a ppc64le edge cluster?

bmpotter (changes requested):

agent-install/agent-install.sh
jacob-rogers: Yes, that was tested on Power platform. For all images I've used private registry, with reassigning paths/tags.
@bmpotter Would you recommend to convert image names in proper way for multi-arch usage to support, or just apply if/else based upon only supported distro/archs here ? I can then fix that.

bmpotter: The `ARCH` variable is set to what is needed for the packages (debs or RPMs) on the current host. I think we need a different variable for `IMAGE_ARCH`, because there is no reason we should have to build both an `amd64` and `x86_64` version of the anax and anax_k8s images. `IMAGE_ARCH` should be set to `amd64` on an amd64 or x86_64 host (regardless of when debian or RHEL) and set to `ppc64le` on Power.

agent-install/agent-install.sh
jacob-rogers: Done

agent-install/agent-install.sh
jacob-rogers: Done, pls recheck. P.S. No global variable was setup there since it shouldn't be reused or redeclared.

agent-install/agent-install.sh
```diff
@@ -1696,13 +1711,23 @@ function get_arch() {
...
 }
 
+# Returns hardware architecture for Docker image names
+function get_image_arch() {
+    local pkg_arch=${ARCH:-$(get_arch)}
+    if [[ $pkg_arch =~ ^(x86_64|amd64)$ ]]; then
+        echo 'amd64'
+    elif [[ $pkg_arch == 'ppc64le' ]]; then
+        echo 'ppc64le'
+    fi
```
bmpotter, line 1728: The problem with this is if a developer adds an arch to the variables at the top of the script, they need to know to also come here and add it. I'd rather have this function only make the modifications needed, instead of needing to be a complete list. Like this:

```
    local image_arch=${ARCH:-$(get_arch)}
    if [[ $image_arch == 'x86_64' ]]; then
        image_arch='amd64'
    elif [[ $image_arch == 'aarch64' ]]; then
        image_arch='arm64'
    fi
    echo $image_arch
```

bmpotter (changes requested):

agent-install/agent-install.sh
jacob-rogers: Ok, fixed that

bmpotter (approved):

dlarson04:
@jacob-rogers   Question for you... did you test the script against RHEL 7.9?  Does your team need support for just RHEL 7.9 on ppc64le or 8.x or both ?

jacob-rogers:
@dlarson04 Only at RHEL 7.9. RedHat 8.x platforms would be great to be supported too, but we didn't run auto install tests there.

dlarson04:
@jacob-rogers I tried doing some testing on a 7.9 ppc64le VM I got from fyre and I hit 2 issues. First

SUPPORTED_REDHAT_VERSION=(8.1 8.2 8.3 $SUPPORTED_REDHAT_VERSION_APPEND) 

7.9 is not listed as supported.... And then, this also failed on my system 

    # Need EPEL for at least jq
    if [[ -z "$(dnf repolist -q epel)" ]]; then
        dnf install -yq https://dl.fedoraproject.org/pub/epel/epel-release-latest-8.noarch.rpm
    fi
    

since this v8 version would not run on my 7.9 system. 

 DId you use a fyre system or something else?

dlarson04:
@jacob-rogers Hey... wondering if you could give me some guidance. I am trying to install docker-ce on a 7.9 or 8.1 system (trying x86 first but then will  try ppc64le). Getting this error

```
[root@severity1 ~]# dnf install -y /root/docker-ce-cli-20.10.4-3.fc33.x86_64.rpm
Updating Subscription Management repositories.
Red Hat Enterprise Linux 8 for x86_64 - Supplementary (RPMs)                                                                                                                                      27 kB/s | 2.1 kB     00:00
Red Hat Enterprise Linux 8 for x86_64 - BaseOS (RPMs)                                                                                                                                             25 kB/s | 2.4 kB     00:00
Red Hat Enterprise Linux 8 for x86_64 - AppStream (RPMs)                                                                                                                                          29 kB/s | 2.8 kB     00:00
Error:
 Problem: conflicting requests
  - nothing provides libc.so.6(GLIBC_2.32)(64bit) needed by docker-ce-cli-1:20.10.4-3.fc33.x86_64
(try to add '--skip-broken' to skip uninstallable packages or '--nobest' to use not only best candidate packages)
```

What is the best way to upgrade to GLIBC_2.32 ?  Thank you

jacob-rogers:
@dlarson04 About docker installation on Power - I had use this guide on RHEL https://developer.ibm.com/components/ibm-power/tutorials/install-docker-on-linux-on-power. Anyway first try to do `yum update` on your machine.
About testing environment - yes, fyre machines were used for that.

PR #2288, opened 2021-02-08
clean up non-UBI anax dockerfiles

### LINKED ISSUE #2256: Clean up non-UBI anax Dockerfiles
- [ ] Remove `anax-in-container/Dockerfile_agbot.amd64` (No need for a non-UBI container, like alpine, for the agbot container in the mgmt hub.)
- [ ] Remove `anax-in-container/Dockerfile.amd64` because that name is misleading (makes it seem like that's the default dockerfile), and because its implementation is very different from the ubi dockerfiles (the former installs the horizon pkgs, the latter just copies the necessary binaries and files into the image)
- [ ] Create a new file called `anax-in-container/Dockerfile.alpine.amd64` that is based on alpine, but the rest of the dockerfile is similar to `anax-in-container/Dockerfile.ubi.amd64`

imskr:
- [X] Remove anax-in-container/Dockerfile_agbot.amd64 (No need for a non-UBI container, like alpine, for the agbot container in the mgmt hub.)

- [X] Remove anax-in-container/Dockerfile.amd64 because that name is misleading (makes it seem like that's the default dockerfile), and because its implementation is very different from the ubi dockerfiles (the former installs the horizon pkgs, the latter just copies the necessary binaries and files into the image)

- [X] Create a new file called anax-in-container/Dockerfile.alpine.amd64 that is based on alpine, but the rest of the dockerfile is similar to anax-in-container/Dockerfile.ubi.amd64

Fixes #2256


imskr:
Thank you @linggao @bmpotter Please review!

bmpotter (approved):

linggao (approved):

PR #2302, opened 2021-02-15
Concurrent map access violation in the Agbot, while caching a service…

MaxMcAdam:
… def

### DIFF
--- a/exchange/cache.go
+++ b/exchange/cache.go
@@ -41,6 +41,27 @@ type CacheEntry struct {
 	Hash        []byte      `json:"hash"`
 }
 
+// Allow getresources to return a copy of cached resource so that multiple threads can use the same resource concurrently
+func (c *CacheEntry) Copy() interface{} {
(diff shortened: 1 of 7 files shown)


exchange/cache.go
```diff
@@ -140,7 +140,13 @@ func GetResourceFromCache(resourceKey string, resourceType string, expirationS u
 	if expirationS > 0 && expired {
 		return nil
 	}
-	return typedEntry.Resource
+	var entryResource interface{}
+	entryResource = typedEntry.Resource
+	if reflect.ValueOf(typedEntry.Resource).Type().Kind() == reflect.Ptr {
+		tempCopy := reflect.Indirect(reflect.ValueOf(typedEntry.Resource))
```
dabooz, line 146: This isnt making a copy of the map in the Resource. The problem is still recreate-able with this fix on.

dabooz (changes requested):

dabooz (approved):

PR #2304, opened 2021-02-15
Added ppc64el support for css and ess Docker images

bencourliss:
Signed-off-by: Ben Courliss <bencourliss@ibm.com>


bencourliss:
In discussions with others, it was determined that PPC is a runtime environment only and therefore this change is not needed.

PR #2317, opened 2021-02-19
Issue 2316 - Image pull does not use the auth for docker.io repo

linggao:

### DIFF
--- a/imagefetch/imagepull.go
+++ b/imagefetch/imagepull.go
@@ -2,7 +2,6 @@ package imagefetch
 
 import (
 	docker "github.com/fsouza/go-dockerclient"
-	"strings"
 
 	"fmt"
 	"github.com/golang/glog"
@@ -10,6 +9,7 @@ import (
 	"github.com/open-horizon/anax/containermessage"
 	"github.com/open-horizon/anax/cutil"
 	"os"
+	"strings"
 	"time"
 )
 
@@ -82,7 +82,6 @@ func pullImageFromRepos(config config.Config, authConfigs map[string][]docker.Au
 			glog.Errorf("Invalid image name format specified: %v", service.Image)
 			return fmt.Errorf("Invalid image name format specified: %v", service.Image)
 		}
-
 		// the image name format is [[repo][:port]/][somedir/]image[:tag][@digest].
 		// tag and digest do not contain '/'
 		if digest != "" {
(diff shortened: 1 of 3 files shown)


imagefetch/imagepull.go
```diff
@@ -116,21 +115,40 @@ func pullImageFromRepos(config config.Config, authConfigs map[string][]docker.Au
...
-				} else if i < len(auth_array)-1 {
-					glog.V(5).Infof("Docker image pull(s) failed for service %v docker image %v with auth name %v. Error: %v. Try next auth.", name, service.Image, auth.Username, err)
-				}
+			domain = "docker.io"
+		}
+
+		// get all the auths for this domain or repo.
+		auth_array := []docker.AuthConfiguration{}
+		for k, _ := range authConfigs {
+			a_auth := authConfigs[k]
```
dabooz, line 126: Why not:
 for k, a_auth := range authConfigs{ ?

imagefetch/imagepull.go
```diff
@@ -116,21 +115,40 @@ func pullImageFromRepos(config config.Config, authConfigs map[string][]docker.Au
...
+		}
+
+		// try auths one at a time
+		var err error
+		for i, auth := range auth_array {
+			err = pullSingleImageFromRepo(client, opts, auth)
+			if err == nil {
+				break
+			} else if i < len(auth_array)-1 {
+				glog.V(5).Infof("Docker image pull(s) failed for service %v docker image %v with auth name %v. Error: %v. Try next auth.", name, service.Image, auth.Username, err)
```
dabooz, line 142: Why not put out a log entry for the last element in the auth_array?

dabooz (changes requested):

imagefetch/imagepull.go
linggao: I worry about the a_auth changes value as we iterates. I forgot what was the trick for this.

dabooz: The problem with iteration occurs when you take the address of an element and try to use it later. In this case you're appending the array element to another array, so I dont think you have that problem. See this example: https://play.golang.org/p/lc8SzA3LVGx that *I think* verifies that there is no problem.

I think you can use authConfig[k] in the append where you use a_auth right now.

dabooz (approved):

PR #2321, opened 2021-02-23
Issue 2292 - Making email optional when creating user

LiilyZhang:


cli/exchange/user.go
```diff
@@ -89,6 +98,14 @@ type UserExchangePatchAdmin struct {
 }
 
 func UserSetAdmin(org, userPwCreds, user string, isAdmin bool) {
+	msgPrinter := i18n.GetMessagePrinter()
+
+	if emailIsOptional, _ := checkExchangeVersionForOptionalUserEmail(org, userPwCreds); emailIsOptional {
```
bmpotter, line 103: Again, i don't think it is important to check `emailIsOptional` here

cli/exchange/user.go
```diff
@@ -104,13 +121,27 @@ func UserSetHubAdmin(org, userPwCreds, user string, isHubAdmin bool) {
...
 		} else {
-			updatedUser := cliutils.UserExchangeReq{Password: existingUser.Password, Admin: existingUser.Admin, HubAdmin: isHubAdmin, Email: existingUser.Email}
-			cliutils.ExchangePutPost("Exchange", http.MethodPut, cliutils.GetExchangeUrl(), "orgs/root/users/"+userId, cliutils.OrgAndCreds(org, userPwCreds), []int{201}, updatedUser, nil)
+			// emailIsOptional is true (exchange version starting from 2.61.0): 
+			// Hub admin users have to be in the root org. We cannot set isHubAdmin to false for a user in root org
+			// root org doesn't have non-hubadmin user
+
+			// emailIsOptional is false (exchange version before 2.61.0):
+			// Can set hubAdmin to desired boolean value
+			if emailIsOptional, _ := checkExchangeVersionForOptionalUserEmail(org, userPwCreds); emailIsOptional {
```
bmpotter, line 134: Again, i don't think it is important to check `emailIsOptional` here

bmpotter (changes requested):

cli/exchange/user.go
```diff
@@ -5,18 +5,21 @@ import (
...
+
 type ExchangeUsers struct {
 	Users     map[string]ExchangeUser `json:"users"`
 	ListIndex int                     `json:"lastIndex"`
 }
 
 type ExchangeUser struct {
 	Password    string `json:"password"`
-	Email       string `json:"email"`
+	Email       string `json:"email,omitempty"`
```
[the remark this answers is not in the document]
LiilyZhang: Set this field to "omitempty", the email field will be omitted from the json encoding. But the email field is still there.
So basically the email field will not present if user has an empty email when using `hzn exchange user list`. 

But I can remove this `omitempty` if you want email to be present even it has empty value for `hzn exchange user list` command.

cli/exchange/user.go
```diff
@@ -66,18 +69,24 @@ func UserCreate(org, userPwCreds, user, pw, email string, isAdmin bool, isHubAdm
...
 
-	if email == "" {
-		if strings.Contains(user, "@") {
-			email = user
-		} else {
-			cliutils.Fatal(cliutils.CLI_INPUT_ERROR, msgPrinter.Sprintf("the email must be specified via -e if the username is not an email address."))
-		}
+	var emailIsOptional bool
+	emailIsOptional, _ = checkExchangeVersionForOptionalUserEmail(org, userPwCreds)
+	fmt.Printf("emailIsOptional: %t\n", emailIsOptional)
```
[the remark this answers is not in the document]
LiilyZhang: Yes, sorry I missed this line when removing my debug statements

cli/exchange/user.go
LiilyZhang: Same as above.`emailIsOptional == true` means user has exchange version later than 2.61.0, then I add my new checks. If `emailIsOptional == false`, it will use the old code path

cli/exchange/user.go
LiilyZhang: This is same as above, use `emailIsOptional` to determine user's exchange version

cli/exchange/user.go
```diff
@@ -66,18 +69,24 @@ func UserCreate(org, userPwCreds, user, pw, email string, isAdmin bool, isHubAdm
...
+		email = user
+	} else if email == "" && !emailIsOptional {
+		cliutils.Fatal(cliutils.CLI_INPUT_ERROR, msgPrinter.Sprintf("the email must be specified via -e if the username is not an email address."))
 	}
 
 	if isHubAdmin && org != "root" {
 		cliutils.Fatal(cliutils.CLI_GENERAL_ERROR, msgPrinter.Sprintf("Only exchange users in the root org can be hubadmins."))
 	}
 
+	if emailIsOptional && !isHubAdmin && org == "root" {
```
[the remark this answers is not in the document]
LiilyZhang: So `emailIsOptional` means user's exchange version is /or later than 2.61.0. I added my new checks only when their exchange has the version with hubadmin fix, to prevent incompatibility.

cli/exchange/user.go
dabooz: I agree with Bruce, dont need the exchange version mixed into the check.

cli/exchange/user.go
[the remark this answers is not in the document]
dabooz: omitempty is ok, since this struct is only used in CLI responses, not exchange requests.

cli/exchange/user.go
dabooz: This path is a little trickier, because a non-hubadmin user could have existed in the exchange before the exchange stopped allowing it. Maybe we want to pretend that isnt possible and ignore the possibility?

dabooz (changes requested):

cli/exchange/user.go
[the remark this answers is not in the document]
LiilyZhang: @dabooz @bmpotter oh I see your point. So this check is also okay for exchange version before 2.61.0. Will remove the version check here

cli/exchange/user.go
```diff
@@ -103,15 +120,9 @@ func UserSetHubAdmin(org, userPwCreds, user string, isHubAdmin bool) {
...
 	if httpCode == 200 {
-		// If the user already exists in the root org, we need to update it with a PUT to set hubAdmin to correct state
-		exchUser := cliutils.AddOrg("root", userId)
-		if existingUser, ok := userList.Users[exchUser]; !ok {
-			cliutils.Fatal(cliutils.NOT_FOUND, msgPrinter.Sprintf("user '%s' not found in list returned from exchange", exchUser))
-		} else {
-			updatedUser := cliutils.UserExchangeReq{Password: existingUser.Password, Admin: existingUser.Admin, HubAdmin: isHubAdmin, Email: existingUser.Email}
-			cliutils.ExchangePutPost("Exchange", http.MethodPut, cliutils.GetExchangeUrl(), "orgs/root/users/"+userId, cliutils.OrgAndCreds(org, userPwCreds), []int{201}, updatedUser, nil)
-		}
+		// Ignore this case if the user is already in root org
```
dabooz, line 125: If the user is in the root org, but wants to unset hubadmin, we need to give them an error message....but think about that for a moment....

I'm thinking that the ability to set hubadmin is no longer interesting. If users in the root org are always hubadmins, then why do we need to set and unset the hubadmin property on the user? A user in the root org cant unset hubadmin, and a user created in the root org is always a hubadmin, so that leaves creation and deletion of the user in the root org.

dabooz (changes requested):

cli/exchange/user.go
```diff
@@ -66,18 +69,26 @@ func UserCreate(org, userPwCreds, user, pw, email string, isAdmin bool, isHubAdm
...
+	var emailIsOptional bool
+	var err error
+	if emailIsOptional, err = checkExchangeVersionForOptionalUserEmail(org, userPwCreds); err != nil {
+		cliutils.Fatal(cliutils.CLI_GENERAL_ERROR, msgPrinter.Sprintf("failed to check exchange version, error: %v", err))
+	}
+
+	if email == "" && strings.Contains(user, "@") {
+		email = user
+	} else if email == "" && !emailIsOptional {
+		cliutils.Fatal(cliutils.CLI_INPUT_ERROR, msgPrinter.Sprintf("the email must be specified via -e if the username is not an email address."))
```
dabooz, line 81: email is a positional parameter, it does not use the -e flag. I think this message should also indicate that email is not optional because the exchange doesnt support it. How about this:
"an email must be specified because the Exchange API version is less than %v.", USER_EMAIL_OPTIONAL_EXCHANGE_VERSION

cli/exchange/user.go
```diff
@@ -66,18 +69,26 @@ func UserCreate(org, userPwCreds, user, pw, email string, isAdmin bool, isHubAdm
...
 	}
 
 	if isHubAdmin && org != "root" {
 		cliutils.Fatal(cliutils.CLI_GENERAL_ERROR, msgPrinter.Sprintf("Only exchange users in the root org can be hubadmins."))
```
dabooz, line 85: Let's remove the word exchange from this message.

cli/exchange/user.go
```diff
@@ -66,18 +69,26 @@ func UserCreate(org, userPwCreds, user, pw, email string, isAdmin bool, isHubAdm
...
 	}
 
+	if !isHubAdmin && org == "root" {
+		cliutils.Fatal(cliutils.CLI_GENERAL_ERROR, msgPrinter.Sprintf("Non-hubadmin user cannot be created in the root org"))
```
dabooz, line 89: Let's reword this to be positive: "Users in the root org must be hubadmins. Omit the -H option."

cli/exchange/user.go
```diff
@@ -89,42 +100,15 @@ type UserExchangePatchAdmin struct {
...
-		// If the user does not exist in the root org and the user isHubAdmin, create them with hubAdmin set to true and no other fields
-		postUserReq := cliutils.UserExchangeReq{HubAdmin: true}
-		httpCode := cliutils.ExchangePutPost("Exchange", http.MethodPost, cliutils.GetExchangeUrl(), "orgs/root/users/"+userId, cliutils.OrgAndCreds(org, userPwCreds), []int{201, 400, 403}, postUserReq, nil)
-		if httpCode == 400 {
-			cliutils.Fatal(cliutils.CLI_GENERAL_ERROR, msgPrinter.Sprintf("Hubadmin status cannot be set because the user %s does not exist in the root org.", user))
-		} else if httpCode == 403 {
-			cliutils.Fatal(cliutils.CLI_GENERAL_ERROR, msgPrinter.Sprintf("Only hubadmin users and the exchange root user can create hubadmins."))
-		}
+	if isAdmin && org == "root" {
+		cliutils.Fatal(cliutils.CLI_GENERAL_ERROR, msgPrinter.Sprintf("User in the root org cannot be set to org admin"))
```
dabooz, line 106: Let's reword this message: A user in the root org cannot be an org admin.

dabooz (changes requested):

cli/exchange/user.go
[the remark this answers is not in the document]
bmpotter: ok, didn't realize that

bmpotter (approved):

dabooz (approved):

Issue #2324, opened 2021-02-24
anax-in-k8s: Dockerfile.ubi only supports amd64


AlexKoe:
The Dockerfile.ubi only supports the amd64 platform since its pulling in a statically linked jq 1.6 binary via wget.

If there is no specific jq 1.6 dependency for that docker image I would suggest to install jq 1.5 from the official Red Hat repositories. 

I have already modified Dockerfile.ubi and will submit a pull request for review.

PR #2325, opened 2021-02-24
Issue 2324, 2337 - arm64 support is missing in anax-in-container and in anax-in-k8s

AlexKoe:
#2324 

The Dockerfile.ubi only supports the amd64 platform since its pulling in a statically linked jq 1.6 binary via wget.

If there is no specific jq 1.6 dependency for that docker image, I would suggest to install jq 1.5 from the official Red Hat repositories. 

I have already modified Dockerfile.ubi and hence submit this pull request for review.


dabooz (changes requested):
Alex, thanks for the contribution. I need you to do 2 things:
1) Did you open issue for this change? If not, I need you to open one and then connect this PR to it. That gives us a forum in which to discuss the issue if needed and we can track it's progress clearly on the zenhub. It's a small price to pay to stay organized. After you have this done, change the name of the PR to "Issue xxxx - some description". The PR title gets into the changelog of the builds so it's easier to see which issues are in a given build.
2) I think you also need to add a dockerfile in ./anax-in-container for arm64 support in the standalone agent container. Or, explain why (and this would be better discussed in the issue you will open) this is only needed for the edge cluster agent.

AlexKoe:
@dabooz Thanks a lot for your comments!

With regards to 1.: I already had opened issue #2324 for that pull request and linked it in my original message above. ;) 

With regards to 2.: I need to look into that one. I had been initially focused on the arm64 K8s support.

AlexKoe:
@dabooz I have added my suggested changes to anax-in-container to this pull request as well. The related issue is #2337

anax-in-container/Dockerfile.ubi.arm64
```diff
@@ -0,0 +1,47 @@
...
+LABEL summary="The agent in a general purpose container."
+LABEL description="A container which holds the edge node agent, to be used in environments where there is no operating system package that can install the agent natively."
+
+ARG DOCKER_VER=19.03.8
+
+# yum is not installed, use microdnf instead
+RUN microdnf update -y --nodocs && microdnf clean all
+
+# add EPEL repo with jq pkg and all deps
+COPY EPEL.repo /etc/yum.repos.d
```
dabooz, line 13: This looks like a copy of the ppc docker file. Their dockerfile is slightly different to pick up jq. Is this what you intended?

dabooz (changes requested):

dabooz:
@AlexKoe Please notice that you have a merge error that you need to resolve. We just merged a ppc change in the same area of the code. I added 2 colleagues to this review, I want to make sure I don't miss something.

anax-in-container/Dockerfile.ubi.arm64
AlexKoe: @dabooz Yes. The ppc version is already picking up jq from UBI repositories. Hence I used that Dockerfile as a template. I only adjusted the docker version to a newer one. I guess sometimes all Dockerfiles in that folder should be potentially consolidated. 😉

AlexKoe:
@dabooz Where/how can I see which change/file caused that merge error?
Thanks in advance!

joewxboy:
[arm/arm64 cluster support for agent](https://trello.com/c/qySLgiqV/6-arm-arm64-cluster-support-for-agent)

dabooz:
@AlexKoe sorry missed your message. Assuming you are in your git workspace on a branch for this PR:
1) git checkout master
2) git pull upstream master   <--- assumes that upstream is the main fork
3) git push origin master   <--- assumes that origin is your fork of anax
4) git checkout <your branch>
5) git rebase master    <--- git will re-apply your PR to your workspace based on the newest main master, if there are conflicts they will be shown in the messages with instructions on what to do next. Usually just need to git add any updated files and then git push -f origin <your branch> to update the PR

Hope this helps

AlexKoe:
@dabooz Sorry for my late reply, but I have been on Easter vacation. I followed your recommend steps and the final step:

git rebase master
Current branch arm64-support is up to date

So it looks to me ok. Please advice, if I might be missing something.  ;)

PR #2339, opened 2021-02-28
Issue 2333 - Modify --device docker command to accept non ':' paths

### LINKED ISSUE #2333: "docker run --device ..." arguments  are not fully supported currently
Docker run's "--device" syntax is "<hostpath>[:<containerpath>][:<accessmode>]" so it accepts things like this:
"/dev/foo:/dev/foo:rwm"
"/dev/foo:/dev/foo"
"/dev/foo"
As equivalent. That is, the access mode argument is ":rwm" (read, write, mknod) when not explicitly specified. And the container path for the device will be the same as the host path when not specified.

Currently ["/dev/foo"] is not allowed in our service definitions (but ["/dev/foo:/dev/foo"] does and this can be used as a workaround until this issue is fixed).

shubhank-saxena:
This fixes #2333

### DIFF
--- a/container/container.go
+++ b/container/container.go
@@ -449,17 +449,23 @@ func (w *ContainerWorker) finalizeDeployment(agreementId string, deployment *con
 		// the cgoup permission can be omitted. It defaults to "rwm" when omitted.
 		for _, givenDevice := range service.Devices {
 			cgp := "rwm"
+			pic := ""
 			sp := strings.Split(givenDevice, ":")
 			if len(sp) == 3 {
 				// the cgroup permission
 				cgp = sp[2]
-			} else if len(sp) < 2 || len(sp) > 3 {
+				pic = sp[1]
+			} else if len(sp) == 2 {
+				pic = sp[1]
+			}else if len(sp) == 1{
+				pic = sp[0]
+			} else if len(sp) <= 0 || len(sp) >3{
 				return nil, fmt.Errorf("Illegal device specified in deployment description: %v", givenDevice)
 			}
 
 			serviceConfig.HostConfig.Devices = append(serviceConfig.HostConfig.Devices, docker.Device{
 				PathOnHost:        sp[0],
-				PathInContainer:   sp[1],
+				PathInContainer:   pic,
 				CgroupPermissions: cgp,
 			})
 		}


container/container.go
```diff
@@ -436,17 +436,25 @@ func (w *ContainerWorker) finalizeDeployment(agreementId string, deployment *con
...
 			cgp := "rwm"
+			pic := ""
 			sp := strings.Split(givenDevice, ":")
 			if len(sp) == 3 {
 				// the cgroup permission
 				cgp = sp[2]
-			} else if len(sp) < 2 || len(sp) > 3 {
+				pic = sp[1]
+			} else if len(sp) == 2 {
+				cgp = cgp
```
linggao, line 446: this line is redundant because you have set it to "rwm" at line 438

container/container.go
```diff
@@ -436,17 +436,25 @@ func (w *ContainerWorker) finalizeDeployment(agreementId string, deployment *con
...
+				pic = sp[1]
+			}else if len(sp) < 1{
+				cgp = cgp
```
linggao, line 449: this line is redundant because you have set it to "rwm" at line 438

container/container.go
```diff
@@ -436,17 +436,25 @@ func (w *ContainerWorker) finalizeDeployment(agreementId string, deployment *con
...
 			sp := strings.Split(givenDevice, ":")
 			if len(sp) == 3 {
 				// the cgroup permission
 				cgp = sp[2]
-			} else if len(sp) < 2 || len(sp) > 3 {
+				pic = sp[1]
+			} else if len(sp) == 2 {
+				cgp = cgp
+				pic = sp[1]
+			}else if len(sp) < 1{
```
linggao, line 448: should be `len(sp) <= 0`

linggao (changes requested):
Please fix the code per review comments and rebase with the master.

container/container.go
shubhank-saxena: Ohh. Added by mistake!

container/container.go
shubhank-saxena: Ohh. Added by mistake!

container/container.go
shubhank-saxena: Ohh. Added by mistake!

container/container.go
linggao: Sorry, my mistake in the comment. The above line should be `else if len(sp) == 1`

container/container.go
```diff
@@ -449,17 +449,23 @@ func (w *ContainerWorker) finalizeDeployment(agreementId string, deployment *con
...
 			if len(sp) == 3 {
 				// the cgroup permission
 				cgp = sp[2]
-			} else if len(sp) < 2 || len(sp) > 3 {
+				pic = sp[1]
+			} else if len(sp) == 2 {
+				pic = sp[1]
+			}else if len(sp) < 1{
+				pic = sp[0]
+			} else if len(sp) <= 0 {
```
linggao, line 462: above line should be `else if len(sp) <= 0 || len(sp) > 3 {`

shubhank-saxena: Oh yes. I understand the issue. Sorry I did not catch this!

linggao (approved):

container/container.go
rajpratyush: @linggao yes that was I wondering about if we will put len(sp)<1 that implies sp is empty then how we can use docker to generate the container it will simply throw an error

Issue #2357, opened 2021-03-05
Have edgeNodeFiles.sh also put agent-install.sh in CSS


bmpotter:
Needed for https://github.com/open-horizon/SDO-support/issues/77 so SDO devices can download `agent-install.sh` from the mgmt hub (instead of having SDO push it to the device, which is really slow).

bmpotter:
Ben's ci/cd pipeline uses `edgeNodeFiles.sh` to push the build artifacts to CSS, and i see that the correct version of `agent-install.sh` is in CSS, so this is working correctly.

PR #2359, opened 2021-03-06
Issue 2357: Have edgeNodeFiles.sh also put agent-install.sh in CSS

bmpotter:
See https://github.com/open-horizon/anax/issues/2357

Signed-off-by: Bruce Potter <bp@us.ibm.com>

### DIFF
--- a/agent-install/edgeNodeFiles.sh
+++ b/agent-install/edgeNodeFiles.sh
@@ -2,6 +2,8 @@
 
 # This script gathers the necessary information and files to install the Horizon agent and register an edge node
 
+function getHznVersion() { local hzn_version=$(hzn version | grep "^Horizon CLI"); echo ${hzn_version##* }; }
+
 # Global constants
 SUPPORTED_NODE_TYPES='ARM32-Deb ARM64-Deb AMD64-Deb x86_64-RPM x86_64-macOS x86_64-Cluster ppc64le-RPM ppc64le-Cluster ALL'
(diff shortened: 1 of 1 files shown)


agent-install/edgeNodeFiles.sh
```diff
@@ -46,6 +48,7 @@ Required Environment Variables:
 
 Optional Environment Variables:
     PACKAGE_NAME: The base name of the horizon content tar file (can include a path). Default: $PACKAGE_NAME
+    HZN_EXCHANGE_USER_AUTH: The exchange credentials to use to publish artifacts in CSS. Must have access to publish to the IBM org. If not set, will use exchange root credentials.
```
bencourliss, line 51: Will the IBM org always exist for people using OpenHorizon outside of IBM?

bencourliss (approved):

agent-install/edgeNodeFiles.sh
bmpotter: Today, yes, the exchange automatically creates it. We are planning to change that, or at least call that org a different name (hopefully in the next release). At that point we'll have to change `edgeNodeFiles.sh`. But right now `edgeNodeFiles.sh` doesn't support pure open-horizon (it uses `cloudctl` and `oc` to access IAM). Hoping to have `edgeNodeFiles.sh` support pure open-horizon in the next release too.

cgiroua (approved):

t-fine (approved):

PR #2361, opened 2021-03-08
Issue #2197 - Remove AgreementTimeoutS and ProtocolTimeoutS from the …

MaxMcAdam:
…default config files


MaxMcAdam:
These variables are now set as configurable scale factors that default to 2x the node max hearbeat interval if it's set or the org-wide max heartbeat interval otherwise

dabooz (approved):

Issue #2358, opened 2021-03-05
Kube worker needs to timeout the custom resource installation if it fails repeatedly


MaxMcAdam:
Otherwise it cannot be uninstalled

PR #2363, opened 2021-03-09
Issue #2358 - Kube worker needs to timeout the custom resource instal…

MaxMcAdam:
…lation if it fails repeatedly

### DIFF
--- a/kube_operator/api_objects.go
+++ b/kube_operator/api_objects.go
@@ -29,7 +29,7 @@ type APIObjectInterface interface {
 // Sort a slice of k8s api objects by kind of object
 // Returns a map of object type names to api object interfaces types, the namespace to be used for the operator, and an error if one occurs
 // Also verifies that all objects are named so they can be found and uninstalled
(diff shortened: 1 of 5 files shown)


kube_operator/api_objects.go
```diff
@@ -588,16 +597,23 @@ func (cr CustomResourceV1) Install(c KubeClient, namespace string) error {
 
 	// the cluster has to create the endpoint for the custom resource, this can take some time
 	// the cr cannot exist without the crd so we don't have to worry about it already existing
-	glog.V(3).Infof(kwlog(fmt.Sprintf("creating operator custom resource %v", cr.CustomResourceObject)))
+	timeout := cr.InstallTimeout
```
dabooz, line 600: I'm trying to get everyone to standardize how config defaults are set. Look at the agent config: https://github.com/open-horizon/anax/blob/master/config/config.go#L297 the default is set into the config object and then the default is overridden when the config file is read in IF the real config file has a value in it.

dabooz (changes requested):

kube_operator/api_objects.go
MaxMcAdam: Fixed

dabooz (approved):

PR #2414, opened 2021-04-06
Issue 2405 - Agent does not respond to the service configstate change…

linggao:
… from the exchange.


agreement/node.go
```diff
@@ -174,8 +171,39 @@ func (w *AgreementWorker) checkNodeUserInputChanges(pDevice *persistence.Exchang
 
 // get the service configuration state from the exchange, check if any of them are suspended.
 // if a service is suspended, cancel the agreements and remove the containers associated with it.
-func (w *AgreementWorker) checkServiceConfigStateChanges(exchDevice *exchange.Device) {
-	glog.V(4).Infof(logString(fmt.Sprintf("Check the service configuration state")))
+func (w *AgreementWorker) checkServiceConfigStateChanges() {
```
dabooz, line 174: Leave the call to checkServiceConfigStateChanges in the checkNodeChanges() function then you don't have to duplicate this code that gets the local device because checkNodeChanges() already does that.

test/gov/service_configstate_test.sh
```diff
@@ -129,22 +137,20 @@ if [ $? -ne 0 ]; then
...
-fi
+read -d '' sconfig <<EOF
+{
+   "url":         "https://bluehorizon.network/services/location",
+   "org":          "e2edev@somecomp.com",
+   "configState": "suspended"
+}
+EOF
+out=$(echo "$sconfig" | curl -sLX POST $CERT_VAR -u $USERDEV_ADMIN_AUTH -H "Content-Type: application/json" -H "Accept: application/json" --data @- ${EXCH_URL}/orgs/userdev/nodes/an12345/services_configstate | jq -r '.')
+echo -e "${PREFIX} $out"
```
dabooz, line 150: You need to check the http status code from the curl so you can fail the test if it's bad.

test/gov/service_configstate_test.sh
```diff
@@ -210,21 +226,18 @@ if [ $? -ne 0 ]; then
...
-fi
+read -d '' sconfig <<EOF
+{
+   "url":         "https://bluehorizon.network/services/location",
+   "org":          "e2edev@somecomp.com",
+   "configState": "active"
+}
+EOF
+out=$(echo "$sconfig" | curl -sLX POST $CERT_VAR -u $USERDEV_ADMIN_AUTH -H "Content-Type: application/json" -H "Accept: application/json" --data @- ${EXCH_URL}/orgs/userdev/nodes/an12345/services_configstate | jq -r '.')
+echo -e "${PREFIX} $out"
```
dabooz, line 237: Check the HTTP status code here too

dabooz (changes requested):

agreement/node.go
linggao: This one has to be taken out of checkNodeChanges() because the change type is `service_configstate` instead of `node`.

test/gov/service_configstate_test.sh
linggao: done

test/gov/service_configstate_test.sh
linggao: done

dabooz (approved):

PR #2416, opened 2021-04-06
Issue 2379 - hzn dev service start -S throws an error for helloworld …

LiilyZhang:
…service

Signed-off-by: zhangl <zhangl@us.ibm.com>

### DIFF
--- a/resource/authentication_manager.go
+++ b/resource/authentication_manager.go
@@ -37,66 +37,80 @@ func (a *AuthenticationManager) GetCredentialPath(key string) string {
 
 // Create a new container authentication credential and write it into the Agent's host file system. The credential file
 // live in a directory named by the key input parameter.
-func (a *AuthenticationManager) CreateCredential(key string, id string, ver string) error {
+func (a *AuthenticationManager) CreateCredential(key string, id string, ver string, secureCreds bool) error {
 	cred, err := GenerateNewCredential(id, ver)
 	if err != nil {
 		return errors.New("unable to generate new authentication token")
 	}
 
-	// The auth folder and auth file need to be access by service account (root or non-root).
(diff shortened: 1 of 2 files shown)


resource/authentication_manager.go
```diff
@@ -37,66 +37,79 @@ func (a *AuthenticationManager) GetCredentialPath(key string) string {
 
 // Create a new container authentication credential and write it into the Agent's host file system. The credential file
 // live in a directory named by the key input parameter.
-func (a *AuthenticationManager) CreateCredential(key string, id string, ver string) error {
+func (a *AuthenticationManager) CreateCredential(key string, id string, ver string, isDev bool) error {
```
dabooz, line 40: The variable isDev, I would prefer that it was named, secureCreds or something like that which communicates that the caller is asking the function to create secure creds or not. The fact that hzn dev uses insecure creds is just one example (happens to be the only example today) where this feature is needed. This comment and the next one also apply to the RemoveCredential function.

resource/authentication_manager.go
```diff
@@ -37,66 +37,79 @@ func (a *AuthenticationManager) GetCredentialPath(key string) string {
...
 	cred, err := GenerateNewCredential(id, ver)
 	if err != nil {
 		return errors.New("unable to generate new authentication token")
 	}
 
+	fileName := path.Join(a.GetCredentialPath(key), config.HZN_FSS_AUTH_FILE)
+	// The following strategy with group permission does not apply to "hzn dev"
```
dabooz, line 47: Please update this comment to describe what's going on more generically. Say something like this: For insecure creds, the group is not needed because the file permissions will be universal.

dabooz (changes requested):

dabooz (approved):

Issue #2451, opened 2021-04-19
Add Vault URL and user creds to agbot configuration


dabooz:
Update the agbot config file (the e2edev agbot and the agbot image built from the anax repo) to include the vault URL and user creds needed for the agbot to access the vault.

PR #2457, opened 2021-04-21
Issue 2451 - Update the agbot config to include vault details

codejaeger:
Signed-off-by: codejaeger <mandaldebabrata123@gmail.com>

Related to #2451 

- [x] Updated agbot config in `/config/config.go` to include vault details
- [x] Added tests to check vault variable enriching from the environment
- [x] Restructured `config_test.go`

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -11,6 +11,7 @@ import (
 
 const ExchangeURLEnvvarName = "HZN_EXCHANGE_URL"
 const FileSyncServiceCSSURLEnvvarName = "HZN_FSS_CSSURL"
+const VaultURLEnvvarName = "HZN_VAULT_ADDR"
 const ExchangeMessageNoDynamicPollEnvvarName = "HZN_NO_DYNAMIC_POLL"
 const OldMgmtHubCertPath = "HZN_ICP_CA_CERT_PATH"
 const ManagementHubCertPath = "HZN_MGMT_HUB_CERT_PATH"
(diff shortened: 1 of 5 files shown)


config/config.go
```diff
@@ -11,6 +11,8 @@ import (
 
 const ExchangeURLEnvvarName = "HZN_EXCHANGE_URL"
 const FileSyncServiceCSSURLEnvvarName = "HZN_FSS_CSSURL"
+const VaultURLEnvvarName = "VAULT_ADDR"
```
dabooz, line 14: openhorizon env vars are always prefixed with HZN_

test/docker/fs/etc/agbot/agbot2.config.tmpl
```diff
@@ -37,6 +37,10 @@
         "CheckUpdatedPolicyS": 15,
         "CSSURL": "${CSS_URL}",
         "CSSSSLCert": "/certs/css.crt",
+        "Vault": {
```
dabooz, line 40: You need to update the ./test/Makefile target agbot-docker-prereqs to set these variables in the config file that e2edev uses at runtime. I'm refering to the for loops in that target which do a textual replacement of css and exchange URLs. We need to do the same for the vault variables you added.

dabooz (changes requested):

config/config.go
codejaeger: @dabooz the variable served one more purpose. To use the vault-cli, this variable needs to be set in the agreement bot.

dabooz (approved):

PR #2487, opened 2021-05-03
Issue 2232 - e2edev improvements: use all-in-1 env

linggao:


test/Makefile
```diff
@@ -93,33 +145,15 @@ endif
...
-DOCKER_CSSDB_CNAME := css-db
-DOCKER_HASHICORP_VAULT_CNAME := vault
-MUL_AGENTS ?= 0
-
-run-agbot: agbot-docker-prereqs test-network
-	touch /tmp/horizon
-	touch /tmp/css.crt
-	@echo "Handling $(DOCKER_AGBOT_CNAME)"
-	docker/docker_run.bash "$(DOCKER_AGBOT_CNAME)" \
+run-test: test-image e2edevtest-docker-prereqs test-network
```
bmpotter, line 153: This comment can be addressed in a follow-on PR: It would be good for the make targets that actually run the tests (and cleanup) be converted to scripts. You need to keep in mind that now many people that aren't core anax developers will be running e2edev tests. Those people will probably be using off-the-shelf versions of most of the containers, and only be building the 1 container they are working on (e.g. exchange), which they may build themselves.

bmpotter (approved):

test/Makefile
linggao: ok, I will put it in the issue 2232 so that we can work on it later.

test/gov/start_mgmt_hub.sh
```diff
@@ -0,0 +1,43 @@
...
+    export START_SECOND_AGBOT=true
+else
+    export START_SECOND_AGBOT=false
+fi
+
+echo -e "${PREFIX} START_SECOND_AGBOT setting is ${START_SECOND_AGBOT}."
+
+cd /tmp
+rm deploy-mgmt-hub.sh
+wget https://raw.githubusercontent.com/open-horizon/devops/master/mgmt-hub/deploy-mgmt-hub.sh
```
dabooz, line 30: In the near future we will need a way for the e2edev user to be able to modify the all in one script and use the local copy on his machine instead of wgetting it.

test/gov/start_mgmt_hub.sh
```diff
@@ -0,0 +1,43 @@
...
+if [ "$MULTIAGBOT" == "1" ]; then
+    export START_SECOND_AGBOT=true
+else
+    export START_SECOND_AGBOT=false
+fi
+
+echo -e "${PREFIX} START_SECOND_AGBOT setting is ${START_SECOND_AGBOT}."
+
+cd /tmp
+rm deploy-mgmt-hub.sh
```
dabooz, line 29: Can we encapsulate this command such that when it fails, we dont get an error message on the screen?

test/Makefile
```diff
@@ -25,66 +22,121 @@ DOCKER_EXCH_TAG := testing
...
-EXCH_TEMPFS := $(CURDIR)/docker/tempfs/exchange
-VAULT_TEMPFS := $(CURDIR)/docker/tempfs/vault
-CSS_CONFIG := $(CURDIR)/docker/fs/etc/edge-sync-service/
-VAULT_CONFIG := $(CURDIR)/docker/fs/etc/vault/config.json
+DOCKER_DEV_OPTS :=  --rm --no-cache --build-arg ARCH=$(arch)
+E2EDEVTEST_TEMPFS := $(CURDIR)/docker/tempfs/e2edevtest
+CERT_LOC ?= 1
+
+# msghub
+MSGHUB_VARS := MSGHUB_BROKER_URL='$(MSGHUB_BROKER_URL)' MSGHUB_API_KEY='$(MSGHUB_API_KEY)'
```
dabooz, line 38: I think we can get rid of this

dabooz (commented):
there's a few empty files, should they be deleted?

test/gov/start_mgmt_hub.sh
linggao: done

test/gov/start_mgmt_hub.sh
linggao: Added as todo to issue 2232.

bmpotter (dismissed):

PR #2504, opened 2021-05-12
Issue 2480 - Change 'hzn register -m <name>' to set the node name for existing nodes

sirotsinskuy:

### DIFF
--- a/cli/register/register.go
+++ b/cli/register/register.go
@@ -193,6 +193,7 @@ func DoIt(org, pattern, nodeIdTok, userPw, inputFile string, nodeOrgFromFlag str
 	exchangePattern := ""
 	httpCode := cliutils.ExchangeGet("Exchange", exchUrlBase, "orgs/"+org+"/nodes/"+nodeId, cliutils.OrgAndCreds(org, nodeIdTok), nil, &devicesResp)
 
+	var existingNodeName string
 	if httpCode != 200 {
 		if userPw == "" {
 			cliutils.Fatal(cliutils.CLI_INPUT_ERROR, msgPrinter.Sprintf("node '%s/%s' does not exist in the Exchange with the specified token, and the -u flag was not specified to provide exchange user credentials to create/update it.", org, nodeId))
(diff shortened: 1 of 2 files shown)


cli/register/register.go
```diff
@@ -193,13 +193,15 @@ func DoIt(org, pattern, nodeIdTok, userPw, inputFile string, nodeOrgFromFlag str
 	exchangePattern := ""
 	httpCode := cliutils.ExchangeGet("Exchange", exchUrlBase, "orgs/"+org+"/nodes/"+nodeId, cliutils.OrgAndCreds(org, nodeIdTok), nil, &devicesResp)
 
+	cliutils.SetWhetherUsingApiKey(userPw)
+	userOrg, userAuth := cliutils.TrimOrg(org, userPw)
```
linggao, line 197: User id and password may not be specified in `hzn register` command.  There are two ways to specify the credentials for this command. Either -u userid:userpasswd or -n nodeid:token, or both.  If the node does not exist, -u must be specified in order to create the node. If the node exists, but do not know the node token, -u must be specified  in order to change the token.  If node exists and you know the token, you can just specify -n for this command.

cli/register/register.go
```diff
@@ -232,10 +235,18 @@ func DoIt(org, pattern, nodeIdTok, userPw, inputFile string, nodeOrgFromFlag str
...
+			existingNodeName = n.Name
 			break
 		}
 	}
 
+	if existingNodeName != "" && existingNodeName != nodeName {
+		msgPrinter.Printf("Updating node name...")
+		msgPrinter.Println()
+		patchNodeReq := cliexchange.NodeExchangePatchName{Name: nodeName}
+		cliutils.ExchangePutPost("Exchange", http.MethodPatch, cliutils.GetExchangeUrl(), "orgs/"+org+"/nodes/"+nodeId, cliutils.OrgAndCreds(userOrg, userAuth), []int{201}, patchNodeReq, nil)
```
linggao, line 247: By now the node id and token is known for all cases, please use node id and token as cred for this call. You need to save the node id and the token for different cases so that they can be used here.

linggao (changes requested):
per comments.

cli/register/register.go
sirotsinskuy: Thanks. Updated creds to use nodeIdTok. Doesn't seem like it needed to be re-saved for different cases - if the token is regenerated - the nodeIdTok value will be updated

linggao (approved):

Issue #2463, opened 2021-04-22
Agbot API to support lifecycle operations on secrets


dabooz:
Add a new resource called /org/<org>/secrets/<vault-secret-name>
GET returns a JSON encoded object indicating the existence or non-existence of the named secret.
200: Secret exists
{
  "exists":<boolean>
}
401: invalid user creds
403: user not in the org specified on the API path
404: Secret does not exist
503: There is no vault component in the management hub. The caller should retry this API call a small number of times with a short delay between calls to ensure that the vault is really not there.

POST/PUT creates or updates a secret. If the secret already exists, it is updated.
Body:
{
  "name":"some/secret/name",
  "secret":"base64 encoded secret bytes"
}
201: created/updated
401: invalid user creds
403: user not in the org specified on the API path
503: here is no vault component in the management hub. The caller should retry this API call a small number of times with a short delay between calls to ensure that the vault is really not there.

DELETE removes a secret.

204: no content, secret is deleted
401: invalid user creds
403: user not in the org specified on the API path
404: secret not found
503: here is no vault component in the management hub. The caller should retry this API call a small number of times with a short delay between calls to ensure that the vault is really not there.

codejaeger:
@dabooz should the status code 503 also be returned when the agbot config does not contain a vault URL. In that case the subsequent message seems irrelevant. So, would it be better of we return a 404 in case of unconfigured vault url.
Also since the user credentials belong to the agbot to authenticate to the vault, where is the org specific information (that the user belongs to this org) is going to come from. I am a bit confused if this will be available from the exchange or the vault itself.

dabooz:
@codejaeger
1) http status code 503 means service unavailable, it is a temporary condition but there is no standard meaning for temporary, could be seconds could be much longer. Since the caller needs to respond differently when a secret is not there 404 vs the vault not being there, having a distinct status code helps the caller.
2) The agbot will sometimes use the callers credentials (called impersonation) when accessing the vault, sometimes its own. A fully formed userid in OH is org/user, and is always carried this way internally.  That is, when a user calls an OH API, the userid in the basic auth header is ALWAYS org/user. Nothing else is supported. This is true for exchange API and for agbot APIs.

PR #2513, opened 2021-05-18
Issue 2463 - Update agbot secrets api

codejaeger:
References issue #2463 

- [x] Implement GET option with tests
- [x] Implement POST/PUT option with tests
- [x] Implement DELETE option with tests
- [ ] Add swagger inline commenting for Agbot secrets API
- [ ] Enrich with inline commenting for relevant sections

### DIFF
--- a/agreementbot/secrets/vault/vault.go
+++ b/agreementbot/secrets/vault/vault.go
@@ -28,10 +28,49 @@ func (vs *AgbotVaultSecrets) String() string {
 	return fmt.Sprintf("Token: %v", vs.token)
 }
 
-func (vs *AgbotVaultSecrets) ListOrgSecret(user, password, org, name string) ([]string, error) {
-	return []string{}, nil
+// This utility will be available to any users within the org.
+func (vs *AgbotVaultSecrets) ListOrgSecret(user, password, org, name string) (map[string]string, error) {
+
+	glog.V(3).Infof(vaultPluginLogString(fmt.Sprintf("list secret %v for org %v", name, org)))
+
+	// Login the user to ensure that the vault ACLs can take effect
+	userVaultToken, err := vs.loginUser(user, password, org)
+	if err != nil {
+		return nil, secrets.ErrorResponse{Msg: fmt.Sprintf("Unable to login user %s, error: %v", user, err), Details: "", RespCode: http.StatusUnauthorized}
+	}
+
+	url := fmt.Sprintf("%s/v1/openhorizon/%s/%s", vs.cfg.GetAgbotVaultURL(), org, name)
+
+	resp, err := vs.invokeVaultWithRetry(userVaultToken, url, http.MethodGet, nil)
+	if resp != nil && resp.Body != nil {
+		defer resp.Body.Close()
+	}
+	if err != nil {
+		return nil, secrets.ErrorResponse{Msg: fmt.Sprintf("Unable to list %s secret %s, error: %v", org, name, err), Details: "", RespCode: http.StatusServiceUnavailable}
+	}
+
+	respBytes, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return nil, secrets.ErrorResponse{Msg: fmt.Sprintf("Unable to read secret %s from %s, error: %v", name, org, err), Details: "", RespCode: http.StatusInternalServerError}
+	} else if httpCode := resp.StatusCode; httpCode == http.StatusNotFound {
+		return nil, secrets.ErrorResponse{Msg: fmt.Sprintf("Secret does not exist. Vault response %s", string(respBytes)), Details: "", RespCode: http.StatusNotFound}
+	} else if httpCode == http.StatusForbidden {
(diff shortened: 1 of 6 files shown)


agreementbot/secrets/vault/vault.go
```diff
@@ -19,59 +20,95 @@ func init() {
...
+	}
+
+	url := fmt.Sprintf("%s/v1/openhorizon/%s/%s", vs.cfg.GetAgbotVaultURL(), org, name)
+
+	resp, err := vs.invokeVaultWithRetry(userVaultToken, url, http.MethodGet, nil)
+	if err != nil {
+		return nil, secrets.ErrorResponse{Msg: fmt.Sprintf("Unable to list %s secret %s, error: %v", org, name, err), Details: "", RespCode: http.StatusServiceUnavailable}
+	}
+
+	defer resp.Body.Close()
```
linggao, line 50: Some times the response body can be non-nil in the error case. So please do the close() after line 45 as 
if resp != nil && resp.Body != nil {
    defer resp.Body.Close()
}

agreementbot/secrets/vault/vault.go
```diff
@@ -19,59 +20,95 @@ func init() {
...
-	}
-
 	respBytes, err := ioutil.ReadAll(resp.Body)
 	if err != nil {
-		return nil, errors.New(fmt.Sprintf("unable to read list %s secrets response, error: %v", org, err))
+		return nil, secrets.ErrorResponse{Msg: fmt.Sprintf("unable to read list %s secrets response, error: %v", org, err), Details: "", RespCode: http.StatusInternalServerError}
+	} else if httpCode := resp.StatusCode; httpCode == http.StatusNotFound {
+		return nil, secrets.ErrorResponse{Msg: "", Details: string(respBytes), RespCode: http.StatusNotFound}
+	} else if httpCode == http.StatusForbidden {
+		return nil, secrets.ErrorResponse{Msg: "", Details: string(respBytes), RespCode: http.StatusForbidden}
```
linggao, line 102: Please add a message here and line 100.

agreementbot/secure_api.go
```diff
@@ -591,9 +592,9 @@ func (a *SecureAPI) secrets(w http.ResponseWriter, r *http.Request) {
 		if vaultSecretName == "" {
 			secretNames, err := a.secretProvider.ListOrgSecrets(ec.GetExchangeId(), ec.GetExchangeToken(), org)
 
-			if err != nil {
-				glog.Errorf(APIlogString(fmt.Sprintf("unable to access secrets provider, error: %v.", err)))
-				writeResponse(w, msgPrinter.Sprintf("unable to access secrets provider, error: %v.", err), http.StatusInternalServerError)
+			if serr, ok := err.(secrets.ErrorResponse); err != nil && ok {
```
linggao, line 595: Should it be serr != nil ?

agreementbot/secure_api.go
```diff
@@ -602,28 +603,19 @@ func (a *SecureAPI) secrets(w http.ResponseWriter, r *http.Request) {
 		} else {
 			secretName, err := a.secretProvider.ListOrgSecret(ec.GetExchangeId(), ec.GetExchangeToken(), org, vaultSecretName)
 
-			if err != nil {
+			if serr, ok := err.(secrets.ErrorResponse); err != nil && ok && (novalue == "" || serr.RespCode != http.StatusNotFound) {
+				glog.Errorf(APIlogString(fmt.Sprintf("unable to access secrets provider, error: %v. %v", serr, serr.Details)))
+				writeResponse(w, msgPrinter.Sprintf("unable to access asda secrets provider, error: %v.", serr), serr.RespCode)
```
linggao, line 608: what is asda?

agreementbot/secrets/vault/vault.go
codejaeger: Hey @linggao , will that be required since `invokeVaultWithRetry` does not introduce a error case where the response body is non nil. Also, I checked the documentation for [httpClient.Do](https://golang.org/pkg/net/http/#Client.Do) and they mention that in the case both are non nil the response body is still closed.

agreementbot/secure_api.go
codejaeger: Yes, it fails when err is nil. There is a test added in `agbot_apitest.sh` which covers that case. As for `serr!=nil` I don't think it matters since the flag `ok` should indicate a failed conversion whether the error is nil or some incompatible type. Anyways, I should change it to `serr!=nil` since that makes more sense in this context.

agreementbot/secure_api.go
codejaeger: Sorry, that was a typo.

agreementbot/secrets/vault/vault.go
linggao: http://devs.cloudimmunity.com/gotchas-and-common-mistakes-in-go-golang/index.html#close_http_resp_body

linggao (approved):

linggao (approved):

PR #2519, opened 2021-05-20
Issue 2517 - Close response body for non-nil body

sirotsinskuy:

### DIFF
--- a/agreementbot/secrets/vault/init.go
+++ b/agreementbot/secrets/vault/init.go
@@ -44,11 +44,12 @@ func (vs *AgbotVaultSecrets) Login() (err error) {
 	}
 
 	resp, err := vs.invokeVaultWithRetry("", url, http.MethodPost, body)
+	if resp != nil && resp.Body != nil {
+		defer resp.Body.Close()
+	}
 	if err != nil {
 		return errors.New(fmt.Sprintf("agbot unable to login, error: %v", err))
 	}
-
-	defer resp.Body.Close()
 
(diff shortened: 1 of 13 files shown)


agreementbot/secrets/vault/init.go
```diff
@@ -48,7 +48,9 @@ func (vs *AgbotVaultSecrets) Login() (err error) {
 		return errors.New(fmt.Sprintf("agbot unable to login, error: %v", err))
 	}
 
-	defer resp.Body.Close()
+	if resp.Body != nil {
+		defer resp.Body.Close()
+	}
```
linggao, line 53: Please move this to after line 47, right after invokeVaultWithRetry because the response body could be non-nil even in the error case.

linggao: The same comment for the rest of the changes.

sirotsinskuy: Updated PR, thanks!

cli/cliutils/cliutils.go
```diff
@@ -1204,6 +1210,9 @@ func InvokeRestApi(httpClient *http.Client, method string, urlPath string, crede
 			http_status := ""
 			if resp != nil {
 				http_status = resp.Status
+				if resp.Body != nil {
+					resp.Body.Close()
+				}
```
linggao, line 1215: Please also add this to line 1226 because it will also exit out.

linggao (approved):

Issue #2520, opened 2021-05-21
typo in "an agent in a container" document


yshk:
In anax/docs/agent_container_manual_deploy.md
https://github.com/open-horizon/anax/blob/master/docs/agent_container_manual_deploy.md
There is trivial typo below.
"hzn register -policy node_pol.json"  --> "hzn register --policy node_pol.json" (not "-" but "--")

lhazuca:
Hi, I would like to fix this. Can you assing me this one?

lhazuca:
Hi, I would like to fix this. Can you assing it to me?

linggao:
@hdiedrich please create a pull request for your fix. thanks.

linggao:
@lhazuca please create a pull request for your fix.

PR #2541, opened 2021-06-03
Fix typo in agent_container_manual_deploy.md

lhazuca:
issue #2520 fixed

### DIFF
--- a/agent-install/agent-install.sh
+++ b/agent-install/agent-install.sh
@@ -64,7 +64,7 @@ ${0##*/} <options>
 Install the Horizon agent on an edge device or edge cluster.
 
 Required Input Variables (via flag, environment, or config file):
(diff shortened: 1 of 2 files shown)


dabooz:
Hi @lhazuca thanks for the contribution. Please make the following updates to the PR:
1) PRs need to be signed off that you are the owner of the contribution. You can do this with the -s flag on git commit.
2) Because of the way the change history tools work, we need to have commits squashed to 1 commit for each PR, otherwise the change history of built packages is very hard to read.
3) Usually we require an issue to be opened for every change, so that a PR can be associated with an issue. Your change is so trivial that I think you can bypass that step this time.
4) Can you tell us a little about yourself, your interest in our project, etc. thanks!

dabooz:
@lhazuca We still need you to sign off the commit with -s, please.

dabooz (approved):

PR #2546, opened 2021-06-03
Issue 2471 - Agent receive secrets from protocol msg, store locally, …

MaxMcAdam:
…cleanup when agreement terminates

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -69,6 +69,7 @@ type Config struct {
 	InitialPollingBuffer             int       // the number of seconds to wait before increasing the polling interval while there is no agreement on the node.
 	MaxAgreementPrelaunchTimeM       int64     // The maximum numbers of minutes to wait for workload to start in an agreement
 	K8sCRInstallTimeoutS             int64     // The number of seconds to wait for the custom resouce to install successfully before it is considered a failure
+	SecretsManagerFilePath           string    // The filepath for the secrets manager to store secrets in the agent filesystem
 
 	// these Ids could be provided in config or discovered after startup by the system
 	BlockchainAccountId        string
@@ -129,7 +130,11 @@ type AGConfig struct {
 // Contains the hashicorp vault configuration used within AGConfig.
 type VaultConfig struct {
 	VaultURL    string // The URL used for accessing the vault.
-	SSLCertPath string // The SSL certificate for the vault
+	SSLCertPath string // The SSL certificate for the vault.
+}
+
+func (c *HorizonConfig) GetSecretsMount() string {
+	return HZN_SECRETS_MOUNT
 }
 
 func (c *HorizonConfig) UserPublicKeyPath() string {
@@ -162,6 +167,15 @@ func (c *HorizonConfig) GetPartitionStale() uint64 {
 
 func (c *HorizonConfig) IsVaultConfigured() bool {
 	return c.AgreementBot.Vault != VaultConfig{}
+}
+
+func (c *HorizonConfig) GetSecretsManagerFilePath() string {
+	secPath := c.Edge.SecretsManagerFilePath
+	if secPath == "" {
+		secPath = "/secrets"
+		secPath = path.Join(getDefaultRunBase(), secPath)
+	}
+	return secPath
 }
 
 func (c *HorizonConfig) GetAgbotCSSURL() string {
@@ -278,6 +292,14 @@ func getDefaultBase() string {
 		basePath = HZN_VAR_BASE_DEFAULT
 	}
 	return basePath
+}
+
+func getDefaultRunBase() string {
+	runBasePath := os.Getenv("HZN_VAR_RUN_BASE")
+	if runBasePath == "" {
+		runBasePath = HZN_VAR_RUN_BASE_DEFAULT
+	}
+	return runBasePath
 }
 
 // some configuration is provided by envvars; in this case we populate this config object from expected envvars
--- a/config/constants.go
+++ b/config/constants.go
@@ -22,6 +22,9 @@ const HTTPIdleConnectionTimeoutS = 120
 
 const HZN_VAR_BASE_DEFAULT = "/var/horizon"
 
+// The default location for ess authentication and secret manager files
+const HZN_VAR_RUN_BASE_DEFAULT = "/var/run/horizon"
+
 // The path to the agent's unix domain socket for the file sync service
 const HZN_FSS_DOMAIN_SOCKET_PATH = "/var/run/horizon"
 const HZN_FSS_DOMAIN_SOCKET = "essapi.sock"
@@ -56,6 +59,9 @@ const HZN_FSS_CERT_KEY_FILE = "key.pem"
 
 // The number of seconds between polls to the CSS for updates.
 const HZN_FSS_POLLING_RATE = 60
+
+// The name of the folder where secrets from the agreement protocol will be stored within a workload container
+const HZN_SECRETS_MOUNT = "/open-horizon-secrets"
 
 // The Default starting exchange message polling interval.
 const ExchangeMessagePollInterval_DEFAULT = 20
--- a/container/container.go
+++ b/container/container.go
@@ -252,6 +252,9 @@ func (w *ContainerWorker) finalizeDeployment(agreementId string, deployment *con
 		// Add a filesystem binding for the FSS (ESS) API SSL client certificate.
 		service.Binds = append(service.Binds, fmt.Sprintf("%v:%v:ro", w.Config.GetESSSSLClientCertPath(), config.HZN_FSS_CERT_MOUNT))
 
+		// Add a filesystem binding for secrets from the agreement protocol to be stored.
+		service.Binds = append(service.Binds, fmt.Sprintf("%v:%v:ro", w.GetSecretsManager().GetSecretsPath(agreementId), config.HZN_SECRETS_MOUNT))
+
 		// Get the group id that owns the service ess auth folder/file. Add this group id in the GroupAdd fields in docker.HostConfig. So that service account in service container can read ess auth folder/file (750)
 		groupAdds := make([]string, 0)
 		if !w.IsDevInstance() {
@@ -531,6 +534,7 @@ type ContainerWorker struct {
 	client            *docker.Client
 	iptables          *iptables.IPTables
 	authMgr           *resource.AuthenticationManager
+	secretMgr         *resource.SecretsManager
 	pattern           string
 	isDevInstance     bool
 	apiServerType     string
@@ -546,6 +550,10 @@ func (cw *ContainerWorker) IsDevInstance() bool {
 
 func (cw *ContainerWorker) GetAuthenticationManager() *resource.AuthenticationManager {
 	return cw.authMgr
+}
+
+func (cw *ContainerWorker) GetSecretsManager() *resource.SecretsManager {
+	return cw.secretMgr
 }
 
 func CreateCLIContainerWorker(config *config.HorizonConfig) (*ContainerWorker, error) {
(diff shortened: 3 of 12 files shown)


main.go
```diff
@@ -156,7 +156,7 @@ func main() {
 	}
 
 	// Initialize the shared authentication manager for service containers to authentication to the agent.
-	authm := resource.NewAuthenticationManager(cfg.GetFileSyncServiceAuthPath())
+	authm := resource.NewAuthenticationManager(cfg.GetFileSyncServiceAuthPath(), "/var/horizon/secrets")
```
dabooz, line 159: This needs to be configurable and overrideable by an env var.

resource/authentication_manager.go
```diff
@@ -17,22 +18,25 @@ import (
 )
 
 type AuthenticationManager struct {
-	AuthPath string
+	CSSAuthPath      string
```
dabooz, line 21: I was envisioning that the secrets would be managed by a separate object, not the auth manager itself. I think we'd like to keep those components separated.

resource/secrets.go
```diff
@@ -0,0 +1,53 @@
...
+import (
+	"encoding/json"
+	"fmt"
+	"github.com/golang/glog"
+	"github.com/open-horizon/anax/containermessage"
+	"github.com/open-horizon/anax/persistence"
+	"path"
+)
+
+func (a AuthenticationManager) HandleNewSerivceSecrets(depDesc *containermessage.DeploymentDescription, agId string) error {
```
dabooz, line 12: So you are conflating persistence of the secret(s) with creating them in the file system in this function. I was envisioning that we'd persist the secrets around the same point where the agreement is persisted, and that container.go would come to this new component to retrieve the secrets and put them into the filesystem. I'm thinking about a case where an agreement proposal is followed closely by a secret update. The secret update should be able to find the secret to be updated in the DB, but it might not be there if its not persisted until container.go.

persistence/secrets.go
```diff
@@ -0,0 +1,78 @@
...
+	"github.com/open-horizon/anax/containermessage"
+)
+
+const SECRETS = "secrets"
+
+func secretKey(agId string, svcName string) string {
+	return fmt.Sprintf("%s/%s", agId, svcName)
+}
+
+func SaveSecrets(db *bolt.DB, secrets map[string]containermessage.Secret, agId string, svcName string) error {
```
dabooz, line 17: You will need an update secret function.
Also you'll need to consider the state that the agent's /secrets API will need in order to know which services a secret has been delivered to.
Finally, I'm not sure that agid is the right key, I think it will have to be service id/name because of the state that the /secret neds to maintain, and it is based on a service not an agreement id.

dabooz (changes requested):

main.go
MaxMcAdam: Fixed

resource/authentication_manager.go
MaxMcAdam: Separated the secrets manager out from auth manager

resource/secrets.go
MaxMcAdam: Done

persistence/secrets.go
MaxMcAdam: Addressed these changes

persistence/secrets.go
```diff
@@ -0,0 +1,112 @@
...
+	}
+}
+
+// Returns the secret from the db if it was there. No error returned if it is not in the db
+func DeleteSecrets(db *bolt.DB, secName string, svcName string) (*PersistedServiceSecret, error) {
+	if sec, err := FindSecrets(db, secName, svcName); err != nil {
+		return nil, err
+	} else {
+		return sec, db.Update(func(tx *bolt.Tx) error {
+			if b, err := tx.CreateBucketIfNotExists([]byte(MICROSERVICE_INSTANCES)); err != nil {
```
dabooz, line 103: Is MICROSERVICE_INSTANCES a typo? :-)

dabooz (changes requested):

config/config.go
```diff
@@ -164,6 +170,14 @@ func (c *HorizonConfig) IsVaultConfigured() bool {
 	return c.AgreementBot.Vault != VaultConfig{}
 }
 
+func (c *HorizonConfig) GetSecretsManagerFilePath() string {
+	path := c.Edge.SecretsManagerFilePath
+	if path == "" {
+		path = "/var/horizon/secrets"
```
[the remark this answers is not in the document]
MaxMcAdam: Change GetSecretsManagerFilePath, I didn't change GetFileSyncServiceAPIListen() since it's default is /var/run/horizon not /var/horizon like HZN_VAR_BASE_DEFAULT

config/constants.go
```diff
@@ -57,6 +57,9 @@ const HZN_FSS_CERT_KEY_FILE = "key.pem"
 // The number of seconds between polls to the CSS for updates.
 const HZN_FSS_POLLING_RATE = 60
 
+// The name of the folder where secrets from the agreement protocol will be stored
```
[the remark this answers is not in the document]
MaxMcAdam: Done

config/config.go
```diff
@@ -69,6 +69,8 @@ type Config struct {
 	InitialPollingBuffer             int       // the number of seconds to wait before increasing the polling interval while there is no agreement on the node.
 	MaxAgreementPrelaunchTimeM       int64     // The maximum numbers of minutes to wait for workload to start in an agreement
 	K8sCRInstallTimeoutS             int64     // The number of seconds to wait for the custom resouce to install successfully before it is considered a failure
+	AgbotURL                         string    // The URL of the agbot secure api.
```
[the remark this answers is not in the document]
MaxMcAdam: Accidentally picked up in a rebase. Removed

persistence/secrets.go
```diff
@@ -0,0 +1,112 @@
+package persistence
```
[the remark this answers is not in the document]
MaxMcAdam: The key is now service org/ service url/ secret name. Then that will return a map of service versions to the secrets. This is because the secrets are coming in with version ranges and need to be findable with individual versions

resource/secrets_manager.go
```diff
@@ -0,0 +1,210 @@
+package resource
```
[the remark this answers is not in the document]
MaxMcAdam: Planning on only having them in the filesystem when they are needed by a container

persistence/secrets.go
```diff
@@ -0,0 +1,112 @@
...
+const SECRETS = "secrets"
+
+type PersistedServiceSecret struct {
+	SvcOrgid        string
+	SvcUrl          string
+	SvcArch         string
+	SvcVersion      string
+	SvcSecretName   string
+	SvcSecretValue  string
+	SvcSecretStatus string
```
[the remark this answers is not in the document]
MaxMcAdam: This is what I am imagining the /secrets API will use to track the state of the updates. I don't know what other values will be needed

persistence/secrets.go
```diff
@@ -0,0 +1,112 @@
...
+	"github.com/open-horizon/anax/policy"
+)
+
+const SECRETS = "secrets"
+
+type PersistedServiceSecret struct {
+	SvcOrgid        string
+	SvcUrl          string
+	SvcArch         string
+	SvcVersion      string
```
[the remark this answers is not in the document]
MaxMcAdam: It looks like it is coming in as a version range so I changed it to that

persistence/secrets.go
```diff
@@ -0,0 +1,112 @@
...
+
+func persistedSecretFromPolicySecret(inputSecrets policy.ServiceSecret) []PersistedServiceSecret {
+	outputSecrets := []PersistedServiceSecret{}
+	for secName, secValue := range inputSecrets.ServiceSecrets {
+		outputSecrets = append(outputSecrets, PersistedServiceSecret{SvcOrgid: inputSecrets.ServiceOrgid, SvcUrl: inputSecrets.ServiceUrl, SvcArch: inputSecrets.ServiceArch, SvcVersion: inputSecrets.ServiceVersion, SvcSecretName: secName, SvcSecretValue: secValue})
+	}
+	return outputSecrets
+}
+
+func secretKey(svcName string, secretName string) string {
```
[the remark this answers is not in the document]
MaxMcAdam: Fixed

persistence/secrets.go
MaxMcAdam: Yep. Fixed

container/container.go
```diff
@@ -1545,6 +1559,11 @@ func (b *ContainerWorker) CommandHandler(command worker.Command) bool {
 			serviceIdentity := cutil.FormOrgSpecUrl(cutil.NormalizeURL(ags[0].RunningWorkload.URL), ags[0].RunningWorkload.Org)
 			sVer := ags[0].RunningWorkload.Version
 
+			if err := b.GetSecretsManager().WriteServiceSecretsToFile(cmd.AgreementLaunchContext.Microservices, agreementId); err != nil {
```
dabooz, line 1562: I think you need this in the ContainerConfigure path also.

dabooz (changes requested):

persistence/secrets.go
```diff
@@ -0,0 +1,281 @@
...
+
+// Find a particular secret, if a version range is provided it must match the exact range on the secret, if a specific version is given return the first matching secret range
+func FindSingleSecretForService(db *bolt.DB, secName string, msDefId string) (*PersistedServiceSecret, error) {
+	allSec, err := FindAllSecretsForMS(db, msDefId)
+	if err != nil {
+		return nil, err
+	}
+
+	if allSec != nil {
+		retSec := allSec.SecretsMap[msDefId]
```
LiilyZhang, line 133: Should this be `retSec := allSec.SecretsMap[secName]`?

MaxMcAdam: Yes, thanks

container/container.go
MaxMcAdam: Moved it to ResourcesCreate() so it runs for both paths

persistence/secrets.go
```diff
@@ -0,0 +1,296 @@
...
+		return fmt.Errorf("Failed to get all secrets for microservice %v. Error was: %v", msDefId, err)
+	} else if allSecsForServiceInDB != nil {
+		secretToSaveAll = *allSecsForServiceInDB
+	}
+
+	if secretToSave.TimeCreated == 0 {
+		secretToSave.TimeCreated = uint64(time.Now().Unix())
+	}
+
+	secretToSave.TimeLastUpdated = uint64(time.Now().Unix())
```
LiilyZhang, line 69: Should this `secretToSave` be `PersistedServiceSecret` or `*PersistedServiceSecret`, since the `TimeCreated` and `TimeLastUpdated` field is updated in the function? 
I tested my secrets API and find after I saved 1 secret, the TimeCreate and TimeLastUpdated will not change even though I passed in different values. Maybe it is related?

test/gov/setup_secrets.sh
```diff
@@ -52,6 +53,7 @@ echo "$RES"
 echo -e "Create netspeed secret2"
 CMD="hzn secretsmanager secret add -o ${USERDEV_ORG} -u ${USERDEV_ADMIN_AUTH} --secretKey ${ORG_SECRET_KEY} -d ${ORG_SECRET_VALUE} ${CREATE_ORG_SECRET2}"
 echo "$CMD"
+<<<<<<< HEAD
```
dabooz, line 56: Merge problem here

container/container.go
```diff
@@ -1700,7 +1714,7 @@ func (b *ContainerWorker) CommandHandler(command worker.Command) bool {
 		sVer := serviceInfo.Version
 
 		// Get the container started
-		if deployment, err := b.ResourcesCreate(lc.Name, "", deploymentDesc, []byte(""), *lc.EnvironmentAdditions, ms_children_networks, serviceIdentity, sVer); err != nil {
+		if deployment, err := b.ResourcesCreate(lc.Name, "", deploymentDesc, []byte(""), *lc.EnvironmentAdditions, ms_children_networks, serviceIdentity, sVer, lc.AgreementIds[0]); err != nil {
```
dabooz, line 1717: We need to discuss this with Ling.

dabooz (changes requested):

dabooz (approved):

PR #2552, opened 2021-06-07
Issue 2524 - hzn unregister command calls systemctl on MacOS

sirotsinskuy:

### DIFF
--- a/cli/unregister/unregister.go
+++ b/cli/unregister/unregister.go
@@ -14,6 +14,7 @@ import (
 	"net/http"
 	"os"
 	"path"
+	"runtime"
 	"strings"
 	"time"
 )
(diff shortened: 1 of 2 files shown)


cli/unregister/unregister.go
```diff
@@ -205,24 +206,40 @@ func DeepClean() error {
...
 	} else {
-		cliutils.Verbose(msgPrinter.Sprintf("Stopping horizon..."))
-		cliutils.RunCmd(nil, "systemctl", "stop", "horizon.service")
-
-		msgPrinter.Printf("Deleting local horizon DB...")
-		msgPrinter.Println()
-		cliutils.RunCmd(nil, "bash", "-c", "rm -f /var/horizon/*.db")
-		cliutils.RunCmd(nil, "bash", "-c", "rm -Rf /etc/horizon/policy.d/*")
+		if runtime.GOOS == "darwin" {
+			outBytes, errBytes := cliutils.RunCmd(nil, "/bin/sh", "-c", "docker ps | grep horizon1")
```
linggao, line 210: `horizon1` is just one container. There could be multiple containers running at the same time: horizon1, horizon2 etc. Need to figure out which container the hzn command is talking about. This can be determined by "HORIZON_URL" env variable (https://github.com/open-horizon/anax/blob/master/cli/cliutils/cliutils.go#L508)  and the port binding of the container.

cli/unregister/unregister.go
```diff
@@ -205,24 +206,40 @@ func DeepClean() error {
...
+			if errBytes != nil && len(errBytes) > 0 {
+				return fmt.Errorf(msgPrinter.Sprintf("Error during checking if anax container is running: %s", errBytes))
+			}
+			if outBytes != nil && len(outBytes) > 0 {
+				cliutils.Verbose(msgPrinter.Sprintf("Restarting horizon container..."))
+				outBytes, errBytes = cliutils.RunCmd(nil, "/bin/sh", "-c", "/usr/local/bin/horizon-container update")
```
linggao, line 216: 'update' might not work here. This is 'deep clean' part, the user usually runs it when the unregistration failed or could not stop so they have to ctrl-c out and run this command with -D. You can add the commands `docker exec -it <horizonx>  rm -f /var/horizon/*.db` and `docker exec -it <horizonx>  rm -Rf /etc/horizon/policy.d/*` before this 'update' command.

cli/unregister/unregister.go
```diff
@@ -205,24 +206,57 @@ func DeepClean() error {
...
+					fmt.Sprintf("docker exec %s rm -f /var/horizon/*.db", containerName))
+				if errBytes != nil && len(errBytes) > 0 {
+					return fmt.Errorf(msgPrinter.Sprintf("Error during removing horizon db files from horizon container: %s", errBytes))
+				}
+				outBytes, errBytes = cliutils.RunCmd(nil, "/bin/sh", "-c",
+					fmt.Sprintf("docker exec %s rm -Rf /etc/horizon/policy.d/*", containerName))
+				if errBytes != nil && len(errBytes) > 0 {
+					return fmt.Errorf(msgPrinter.Sprintf("Error during removing policy files from horizon container: %s", errBytes))
+				}
+				outBytes, errBytes = cliutils.RunCmd(nil, "/bin/sh", "-c", "/usr/local/bin/horizon-container update")
```
linggao, line 233: For horizon2, for example, the command should be `/usr/local/bin/horizon-container update 2`. Otherwise, it will update horizon1 by default.

linggao (changes requested):
per comment.

cli/unregister/unregister.go
sirotsinskuy: Updated, thanks

linggao (approved):

PR #2571, opened 2021-06-11
Issue 2570 - Added additional e2edev MMS tests

### LINKED ISSUE #2570: Add e2edev test for MMS publishing of large files
The e2edev tests don't test publishing of large files to CSS. This needs to be tested due to a failure that occurred in testing against an OpenShift deployed IEAM cluster when uploading a 1GB file. The underlying ingress used by the OpenShift env was not tuned properly for large file uploads and hit OOMKilled termination status and pod restarts when doing `hzn publish`.

bencourliss:
Resolves #2570

### DIFF
--- a/test/gov/clean_css.sh
+++ b/test/gov/clean_css.sh
@@ -14,5 +14,18 @@ hzn mms object delete -t "model" -i "multires.tgz"
 hzn mms object delete -t "model" -i "policy-basicres.tgz"
 hzn mms object delete -t "model" -i "policy-multires.tgz"
 sleep 5
+hzn mms object delete -t "test" -i "test-medium1"
+hzn mms object delete -t "test" -i "test-large1"
+hzn mms object delete -t "test" -i "test1"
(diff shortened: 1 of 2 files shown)


test/gov/clean_css.sh
```diff
@@ -14,5 +14,18 @@ hzn mms object delete -t "model" -i "multires.tgz"
 hzn mms object delete -t "model" -i "policy-basicres.tgz"
 hzn mms object delete -t "model" -i "policy-multires.tgz"
 sleep 5
+hzn mms object delete -t "test" -i "test-medium1"
```
LiilyZhang, line 17: Just a general question about this class: where is this script called? I didn't find the place. Are we still using this class?

bencourliss: Good catch, I just did a search for where we were using `hzn mms object` and this was one of the files that came back. I didn't think to check where it's being called from and it doesn't appear to be used anywhere. I did use it to clean things up when I was testing though.

PR #2576, opened 2021-06-15
Issue 2393 - hzn dev service log should have an argument for the URL of the service to display

Isabel-Gan:
Signed-off-by: Isabel Gan <Isabel.Gan@ibm.com>

### DIFF
--- a/cli/hzn.go
+++ b/cli/hzn.go
@@ -3,10 +3,13 @@ package main
 
 import (
 	"flag"
-	"github.com/open-horizon/anax/cli/sdo"
-	"github.com/open-horizon/anax/version"
 	"os"
 	"strings"
+
+	"github.com/open-horizon/anax/cli/sdo"
+	"github.com/open-horizon/anax/version"
+
+	"runtime"
 
 	"github.com/open-horizon/anax/cli/agreement"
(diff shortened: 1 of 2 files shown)


cli/hzn.go
```diff
@@ -232,7 +234,7 @@ Environment Variables:
 
 	devServiceCmd := devCmd.Command("service", msgPrinter.Sprintf("For working with a service project."))
 	devServiceLogCmd := devServiceCmd.Command("log", msgPrinter.Sprintf("Show the container/system logs for a service."))
-	devServiceLogCmdServiceName := devServiceLogCmd.Flag("service", msgPrinter.Sprintf("(DEPRECATED) This flag is deprecated and is replaced by -c.")).Short('s').String()
```
linggao, line 235: Still need to keep this line so the old user will know.

linggao, line 235: Still need to keep this line so that the old user will know what happened to this flag.  Since `devServiceLogCmdServiceName` is used next line, you can just do `devServiceLogCmd.Flag("service", ...)` without assigning it to any variable.

Isabel-Gan: addressed in most recent changes

linggao (approved):

PR #2577, opened 2021-06-15
Issue 2567 - Allow user to specify public key and private key for MMS object data signing

Polber:
Signed-off-by: Jeff Kinard <jeff@thekinards.com>

### DIFF
--- a/cli/hzn.go
+++ b/cli/hzn.go
@@ -570,6 +570,7 @@ Environment Variables:
 	mmsObjectPublishSkipIntegrityCheck := mmsObjectPublishCmd.Flag("noIntegrity", msgPrinter.Sprintf("The publish command will not perform a data integrity check on the uploaded object data. It is mutually exclusive with --hashAlgo and --hash")).Bool()
 	mmsObjectPublishDSHashAlgo := mmsObjectPublishCmd.Flag("hashAlgo", msgPrinter.Sprintf("The hash algorithm used to hash the object data before signing it, ensuring data integrity during upload and download. Supported hash algorithms are SHA1 or SHA256, the default is SHA1. It is mutually exclusive with the --noIntegrity flag")).Short('a').String()
 	mmsObjectPublishDSHash := mmsObjectPublishCmd.Flag("hash", msgPrinter.Sprintf("The hash of the object data being uploaded or downloaded. Use this flag if you want to provide the hash instead of allowing the command to automatically calculate the hash. The hash must be generated using either the SHA1 or SHA256 algorithm. The -a flag must be specified if the hash was generated using SHA256. This flag is mutually exclusive with --noIntegrity.")).String()
+	mmsObjectPublishPrivKeyFile := mmsObjectPublishCmd.Flag("private-key-file", msgPrinter.Sprintf("The path of a private key file to be used to sign the model.")).Short('k').ExistingFile()
 	mmsStatusCmd := mmsCmd.Command("status", msgPrinter.Sprintf("Display the status of the Horizon Model Management Service."))
 
 	nodeCmd := app.Command("node", msgPrinter.Sprintf("List and manage general information about this Horizon edge node."))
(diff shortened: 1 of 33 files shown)


Polber:
The user now has the ability to specify a -k flag to utilize a custom private key. Since the previous implementation always generated a new RSA key pair to sign the object, that has been done here. The user is expected to generate a private/public RSA key pair, but since the public key is generated from the private key automatically in the code, the user does not need to specify the public key file when running the command - that is why only the private key needs to be included. However, when comparing this command to the `hzn exchange service publish` command which also supports custom key files, there are some notable differences that led me to the following questions:

Should there be a default file path and/or default environment variable pointing to the file(s) to be used, similar to how `hzn exchange service publish` works (something like HZN_MMS_PRIVATE_KEY_FILE)?

Should the public key be a certificate like it is for `hzn exchange service publish` or just a regular RSA public key like what is generated when no key is specified?

Should the public key still be specified and compared to the generated public key to ensure they are the same?

Polber:
It now will try to use a private key file in the following order:
1. -k flag
2. HZN_PRIVATE_KEY_FILE env variable
3. $HOME/.hzn/keys/service.private.key
4. /tmp/keys/.hzn/keys/service.private.key

This is the same check order that `hzn exchange service publish` uses (Although I think the fourth may be a mistake that needs refactoring for both commands - should not include /.hzn/keys in path)

If it can't find a file using any of the above, it will generate a key in the code.

There is no need to specify a public key file as it is generated from the private key, so the -K flag has not been added.

LiilyZhang:
@Polber Could you also add/update e2edev test cases? The `hzn mms object publish` test is in `anax/test/gov/sync_service_test.json`

Polber:
I added tests to ensure the correct private/public keys were published with the object for the following 3 cases:
1. Specify with -k flag
2. Set HZN_PRIVATE_KEY_FILE environment variable 
    * -k flag not given
3. Use default keys in ~/.hzn/keys/ 
    * -k flag not given and HZN_PRIVATE_KEY_FILE not set

cli/hzn.go
```diff
@@ -571,6 +571,7 @@ Environment Variables:
...
 	mmsObjectPublishDSHashAlgo := mmsObjectPublishCmd.Flag("hashAlgo", msgPrinter.Sprintf("The hash algorithm used to hash the object data before signing it, ensuring data integrity during upload and download. Supported hash algorithms are SHA1 or SHA256, the default is SHA1. It is mutually exclusive with the --noIntegrity flag")).Short('a').String()
 	mmsObjectPublishDSHash := mmsObjectPublishCmd.Flag("hash", msgPrinter.Sprintf("The hash of the object data being uploaded or downloaded. Use this flag if you want to provide the hash instead of allowing the command to automatically calculate the hash. The hash must be generated using either the SHA1 or SHA256 algorithm. The -a flag must be specified if the hash was generated using SHA256. This flag is mutually exclusive with --noIntegrity.")).String()
+	mmsObjectPublishPrivKeyFile := mmsObjectPublishCmd.Flag("private-key-file", msgPrinter.Sprintf("The path of a private key file to be used to sign the model. If not specified, the environment variable HZN_PRIVATE_KEY_FILE will be used. If none of them are set, ~/.hzn/keys/service.private.key is the default.")).Short('k').ExistingFile()
```
dabooz, line 574: Please reword the help as follows:
The path of a private key file to be used to sign the object. The corresponding public key will be stored in the MMS to ensure integrity of the object. If not specified, the environment variable HZN_PRIVATE_KEY_FILE will be used to find a private key. If not set, ~/.hzn/keys/service.private.key will be used. If it does not exist, an RSA key pair is generated only for this publish operation and then the private key is discarded.

dabooz (changes requested):
There is a .DS_Store binary file in this PR. You might need to add that to git ignore if it's a leftover from running tests

dabooz (approved):

PR #2598, opened 2021-06-28
Issue 2477 - Update hzn deploycheck command  and agbot API to check for secret compa…

linggao:
…tibility


agreementbot/agreementworker.go
```diff
@@ -651,8 +638,12 @@ func (b *BaseAgreementWorker) InitiateNewAgreement(cph ConsumerProtocolHandler,
 		} else {
 
 			foundWorkload = true
-			svcIds = make([]string, len(sIds))
-			copy(svcIds, sIds)
+
+			// copy all the service ids
```
dabooz, line 642: Can you use the built in copy function here?

dabooz (changes requested):
Please rebase it on current master so I can run some tests on it.

agreementbot/agreementworker.go
linggao: The depServices is a map, we just need its keys in svcIds.

dabooz (approved):

PR #2604, opened 2021-06-28
Issue 2569 - Panic in agbot when mms objects deleted

dlarson04:

### DIFF
--- a/agreementbot/object_manager.go
+++ b/agreementbot/object_manager.go
@@ -195,14 +195,19 @@ func (m *MMSObjectPolicyManager) UpdatePolicies(org string, updatedPolicies *exc
 		glog.V(5).Infof(mmsLogString(fmt.Sprintf("Starting object policy garbage collection")))
 		for org, serviceMap := range m.orgMap {
 			for service, peList := range serviceMap {
-				for ix, pe := range peList {
+				// Since we can't modify the policy entry list while iterating over it, collect the entries that we want to keep and add them back into the object policy manager once the iteration is complete.
+				finalList := make([]MMSObjectPolicyEntry, 0)
+				for _, pe := range peList {
 					if obj, err := objQueryHandler(pe.Policy.OrgID, pe.Policy.ObjectID, pe.Policy.ObjectType); err != nil {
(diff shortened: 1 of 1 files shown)


agreementbot/object_manager.go
```diff
@@ -195,14 +195,18 @@ func (m *MMSObjectPolicyManager) UpdatePolicies(org string, updatedPolicies *exc
 		glog.V(5).Infof(mmsLogString(fmt.Sprintf("Starting object policy garbage collection")))
 		for org, serviceMap := range m.orgMap {
 			for service, peList := range serviceMap {
-				for ix, pe := range peList {
+				finalList := make([]MMSObjectPolicyEntry, 0)
```
dabooz, line 198: Please add a comment before this line, something like:
// Since we cant modify the policy entry list while iterating over it, collect the entries that we want to keep and add them back into the object policy manager once the iteration is complete.

And same comment in the next section too.

dabooz (changes requested):

dabooz (dismissed):

MaxMcAdam (dismissed):

dabooz:
@dlarson04 The PR is failing the e2edev tests because there is a bug, which we missed in the code review. Here is a diff that fixes two different problems:

$ git diff
diff --git a/agreementbot/object_manager.go b/agreementbot/object_manager.go
index c0e25f75..d44cbd76 100644
--- a/agreementbot/object_manager.go
+++ b/agreementbot/object_manager.go
@@ -242,9 +242,11 @@ func (m *MMSObjectPolicyManager) UpdatePolicies(org string, updatedPolicies *exc
                                                policyReplaced = pe.Policy
                                                glog.V(3).Infof(mmsLogString(fmt.Sprintf("object %v/%v %v policy removed from %v cache.", objPol.OrgID, objPol.ObjectID, objPol.ObjectType, service)))
                                        }
+                               } else {
+                                       finalList = append(finalList, pe)
                                }
                        }
-                       m.orgMap[org][service] = finalList
+                       m.orgMap[objPol.OrgID][service] = finalList
                }
 
                // Now run through each service in the updated policy and figure out if there is a change or if it's new.

Please update the PR with this change.

dabooz (approved):

PR #2606, opened 2021-06-29
Issue2605 - MMS download needs to stream object data into a file inst…

LiilyZhang:
…ead of save in memory before writing to a file

Signed-off-by: zhangl <zhangl@us.ibm.com>

### DIFF
--- a/cli/sync_service/download.go
+++ b/cli/sync_service/download.go
@@ -1,7 +1,6 @@
 package sync_service
 
 import (
-	"bytes"
 	"crypto/rsa"
 	"crypto/x509"
 	"encoding/base64"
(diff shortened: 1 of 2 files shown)


cli/sync_service/download.go
```diff
@@ -109,23 +90,49 @@ func ObjectDownLoad(org string, userPw string, objType string, objId string, fil
...
+			}
+			msgPrinter.Println("Verifying digital signature is done.")
+			rename = true
+		}
 
-	if _, err := file.Write(data); err != nil {
-		cliutils.Fatal(cliutils.INTERNAL_ERROR, msgPrinter.Sprintf("Failed to save data for object '%s' of type '%s' to file %s", objId, objType, fileName))
+	}
+
+	if rename {
```
dabooz, line 117: Can you move this rename logic into the VerifyDataSig function?

cli/sync_service/download.go
```diff
@@ -109,23 +90,49 @@ func ObjectDownLoad(org string, userPw string, objType string, objId string, fil
...
+		// rename fileName.tmp
+                tmpFileName := fmt.Sprintf("%s.tmp", fileName)
+                if err := os.Rename(tmpFileName, fileName); err != nil {
+                        cliutils.Fatal(cliutils.INTERNAL_ERROR, msgPrinter.Sprintf("Failed to rename file '%s' to '%s', err: %v", tmpFileName, fileName, err))
+                }
+	} else {
```
dabooz, line 123: This else clause would become the else of the if !skipDigitalVerify

dabooz (changes requested):

cli/sync_service/download.go
LiilyZhang: Yes, that is right. So 
if !skipDigitalVerify, 
     the data will be streamed to a tmp file, at same time generate hash + verify.
then rename the tmp file
else 
    the data is stream to the file directly

LiilyZhang: @dabooz ah I think I know what you mean. Yea after moving the rename logic, will also change this part

cli/sync_service/download.go
```diff
@@ -157,7 +168,56 @@ func VerifyDataSig(dataReader io.Reader, publicKey string, signature string, has
...
+}
+
+func (wc *WriteCounter) Write(p []byte) (n int, err error) {
+	n, err = wc.PassInWriter.Write(p)
+	wc.ChunkNumber += 1
+	wc.PrintProgress(n)
+	return n, err
+}
+
+func (wc *WriteCounter) PrintProgress(n int) {
```
dabooz, line 216: I really dont like how the dots work. It would look better if we could use the % counter like we do for upload.

dabooz (changes requested):

dabooz (approved):

PR #2610, opened 2021-06-30
Issue 2609 - Optimization: Make sure all patterns/policies are search…

dlarson04:
Keep map in NodeSearch of patterns/policies completely searched. This allows us to iterate through all policies and patterns in search before repeating search of patterns/policies already searched. This improves agreement times for scenario where hundreds or thousands of agents register with node policies that match multiple Deployment Policies

### DIFF
--- a/agreementbot/node_search.go
+++ b/agreementbot/node_search.go
@@ -11,8 +11,10 @@ import (
 	"github.com/open-horizon/anax/externalpolicy"
 	"github.com/open-horizon/anax/policy"
 	"sync"
+	"strings"
 	"time"
 )
+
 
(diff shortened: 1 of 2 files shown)


dabooz:
@dlarson04 Did a number of scaling tests on this code to ensure that:
1) It doesn't affect small scale/demo environments. Agreements are still completed fairly quickly.
2) While some scale scenarios perform slower (i.e. some nodes take longer to get into agreement) than they used to, the overall experience of how quickly agreements can be completed is more balanced than it was before resulting is better overall response times to get into an agreement.
3) New nodes being targetted by new policies that arrive in the system during heavy agreement creation are not lost.

dabooz (approved):

PR #2615, opened 2021-07-01
Issue 2575 - Obscure secret details when displaying proposals

Polber:
Signed-off-by: polber <jeff@thekinards.com>

* Agbot and Agent API’s now return the agreement with the proposal edited to obscure the secret details
* Agreement GET API throws Error if not in correct format because it will be unable to unmarshal and marshal the data correctly: 
* Writes to logs when secret details are not obscured correctly


agreementbot/api.go
```diff
@@ -442,6 +446,31 @@ func (a *API) agreement(w http.ResponseWriter, r *http.Request) {
...
+	} else if err := json.Unmarshal([]byte(tsandcs["tsandcs"].(string)), &tsandcs); err != nil {
+		return err
+	} else if _, ok := tsandcs["secretDetails"]; !ok {
+		// there are no secrets to obscure, so return nil
+		return nil
+	}
+	secretDetails := tsandcs["secretDetails"].([]interface{})
+
+	// loop over secrets, if any
+	for _, secretDetail := range secretDetails {
```
dabooz, line 465: As part of my PR for issue 2475 I had to rework the data structure for secretDetails. You will have to change this code to adapt to the new changes. My previous comment should make this easier.

agreementbot/api.go
```diff
@@ -442,6 +446,31 @@ func (a *API) agreement(w http.ResponseWriter, r *http.Request) {
 	}
 }
 
+// Obscure the secret details for all secrets in the given proposal
+func obscureSecretDetails(p *persistence.Agreement) error {
+	prop := p.Proposal
+
+	var tsandcs map[string]interface{}
+	if err := json.Unmarshal([]byte(prop), &tsandcs); err != nil {
```
dabooz, line 454: There is a function for demarshalling proposals already: abstractprotocol.DemarshalProposal(proposal), and there's a function for demarshalling the TsAndCs: policy.DemarshalPolicy(proposal.TsAndCs()). I would put the code to obscure the secrets in ./policy/policy_file.go. The Policy type represents the internal policy objects that Patterns and Deployment Policies get converted to.

api/path_agreement.go
```diff
@@ -44,6 +49,32 @@ func FindAgreementsForOutput(db *bolt.DB) (map[string]map[string][]persistence.E
 	return wrap, nil
 }
 
+// Obscure the secret details for all secrets in the given proposal
+func obscureSecretDetails(p *persistence.EstablishedAgreement) error {
```
dabooz, line 53: And you can use the code from the Policy type in both the agent and the agbot, so you wont have to duplicate it here.

agreementbot/api.go
```diff
@@ -387,6 +388,9 @@ func (a *API) agreement(w http.ResponseWriter, r *http.Request) {
 				} else {
 
 					for _, agreement := range ags {
+						if err := obscureSecretDetails(&agreement); err != nil {
+							glog.Warning(APIlogString(fmt.Sprintf("failed to obscure secret details, error: %v", err)))
```
dabooz, line 392: You missed the /agreement/{id} section right before this code, it needs to be updated too.

dabooz (changes requested):

agreementbot/api.go
```diff
@@ -365,6 +366,8 @@ func (a *API) agreement(w http.ResponseWriter, r *http.Request) {
 				http.Error(w, "Internal server error", http.StatusInternalServerError)
 			} else if ag == nil {
 				writeInputErr(w, http.StatusBadRequest, &APIUserInputError{Input: "id", Error: "agreement id not found"})
+			} else if err := abstractprotocol.ObscureProposalSecret(&ag.Proposal); err != nil {
```
dabooz, line 369: Modify the agreement.Proposal here instead of within the obscure function.

agreementbot/api.go
```diff
@@ -365,6 +366,8 @@ func (a *API) agreement(w http.ResponseWriter, r *http.Request) {
...
+				glog.Error(APIlogString(fmt.Sprintf("failed to obscure secret details, error: %v", err)))
 			} else {
```
dabooz, line 371: You need to obscure the agreement.Policy object also. It has the same secretDetails in it. This is only a problem on the agbot side, the agent only has the proposal.

dabooz (changes requested):

dabooz (approved):

PR #2620, opened 2021-07-02
Issue 2617 - Panic from agbot api after call to /health

MaxMcAdam:


MaxMcAdam:
The change in agreementbot/secrets/vault/http.go is for another panic I hit while trying to recreate the issue

dabooz (approved):

PR #2633, opened 2021-07-12
Issue 2596 - Agent fails to verify signature due to internal error

dlarson04:

- Add uniqueness to downloaded signing key name to avoid 2 keys overwriting each other
- Change order of validating proposal to rule out proposal for other reasons before checking keys

### DIFF
--- a/governance/microservice.go
+++ b/governance/microservice.go
@@ -40,6 +40,23 @@ func (w *GovernanceWorker) governMicroservices() int {
 	return 0
 }
 
+// Since we changed to saving the signing key with the agreement id, we need to make sure we delete the key when done with it
+// to avoid filling up the filesystem
+func (w *GovernanceWorker) cleanupSigningKeys(keys []string) {
+
+        errHandler := func(keyname string) api.ErrorHandler {
+                return func(err error) bool {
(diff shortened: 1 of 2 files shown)


dabooz (approved):

governance/microservice.go
```diff
@@ -132,18 +153,29 @@ func (w *GovernanceWorker) StartMicroservice(ms_key string, agreementId string,
 							fn = fmt.Sprintf("%v.pem", key)
 						}
 
-						api.UploadPublicKey(fn, []byte(content), w.Config, errHandler(fn))
+						// Keys for different services might have the same key name like service.public.pem so prepend something unique like the agreement id 
+						// but then we have to make sure we delete the key when done with it
+						prepend_string := agreementId + "_" + strconv.Itoa(num_signing_keys) + "_"
```
linggao, line 158: StartMicroservice get called with agreementId as an empty string in some cases.  If agreementId is empty, please use ms_key instead.

linggao (changes requested):
per comment

governance/microservice.go
dlarson04: changed per comment

linggao (approved):

PR #2638, opened 2021-07-13
Issue 2612 - hzn secretsmanager secret read command that allows authenticated users to see secret details

Isabel-Gan:
Signed-off-by: Isabel Gan <Isabel.Gan@ibm.com>

API changes: `LIST` method supported for listing multiple and individual secrets (for an individual secret, will return `{ exists: true/false }`. `GET` is supported only for individual secrets, and will return the secret details if the user is allowed to read the secret.

### DIFF
--- a/agreementbot/agreementworker.go
+++ b/agreementbot/agreementworker.go
@@ -852,7 +852,7 @@ func (b *BaseAgreementWorker) ValidateAndExtractSecrets(consumerPolicy *policy.P
 				}
 
 				// Call the secret manager plugin to get the secret details.
-				details, err := b.secretsMgr.GetSecretDetails(exchange.GetOrg(deviceId), secretUser, shortSecretName)
+				details, err := b.secretsMgr.GetSecretDetails(b.GetExchangeId(), b.GetExchangeToken(), exchange.GetOrg(deviceId), secretUser, shortSecretName)
 
 				if err != nil {
 					glog.Errorf(BAWlogstring(workerId, fmt.Sprintf("error retrieving secret %v for policy %v, service %v/%v %v, error: %v", secretName, consumerPolicy.Header.Name, binding.ServiceOrgid, binding.ServiceUrl, binding.ServiceVersionRange, err)))
--- a/agreementbot/governance.go
+++ b/agreementbot/governance.go
@@ -251,7 +251,7 @@ func (w *AgreementBotWorker) GovernAgreements() int {
 											continue
 										}
 
-										details, err := w.secretProvider.GetSecretDetails(ag.Org, secretUser, secretName)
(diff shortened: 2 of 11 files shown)


agreementbot/agreementworker.go
```diff
@@ -852,7 +852,7 @@ func (b *BaseAgreementWorker) ValidateAndExtractSecrets(consumerPolicy *policy.P
 				}
 
 				// Call the secret manager plugin to get the secret details.
-				details, err := b.secretsMgr.GetSecretDetails(exchange.GetOrg(deviceId), secretUser, shortSecretName)
+				details, err := b.secretsMgr.GetSecretDetails(b.config.AgreementBot.ExchangeId, b.config.AgreementBot.ExchangeToken, exchange.GetOrg(deviceId), secretUser, shortSecretName)
```
dabooz, line 855: I think the more correct way is this:
GetSecretDetails(b.GetExchangeId(), b.GetExchangeToken(), ...)

agreementbot/governance.go
```diff
@@ -251,7 +251,7 @@ func (w *AgreementBotWorker) GovernAgreements() int {
 											continue
 										}
 
-										details, err := w.secretProvider.GetSecretDetails(ag.Org, secretUser, secretName)
+										details, err := w.secretProvider.GetSecretDetails(w.EC.Id, w.EC.Token, ag.Org, secretUser, secretName)
```
dabooz, line 254: I think the more correct way is this:
GetSecretDetails(w.GetExchangeId(), w.GetExchangeToken(), ...)

cli/cliutils/cliutils.go
```diff
@@ -876,6 +876,89 @@ func AgbotGet(urlSuffix, credentials string, goodHttpCodes []int, structure inte
...
+	}
+
+	respBody := io.Reader(resp.Body)
+	if resp.Header.Get("Content-type") == "application/octet-stream" {
+		// Show progress of binary files downloading
+		msgPrinter.Print("Downloading object")
+		chunkNumber := 0
+		respBody = &progressReader{resp.Body, func(r int) {
+			chunkNumber++
+			if r > 0 {
```
dabooz, line 912: Why do we need this for LIST? The response will never be very big. I expected this function to look a  lot more like the GET.

dabooz (changes requested):

cli/cliutils/cliutils.go
Isabel-Gan: I just got rid of the progress-checking code; my code for AgbotGet uses ExchangeGet and there wasn't an existing ExchangeList so I took the code from ExchangeGet and switched out GET for LIST

dabooz (approved):

PR #2674, opened 2021-07-27
Issue 2624 - Check agbot and agent logs for secret details

Polber:
_abstractprotocol/protocol.go_ `CreateProposal()` function was updated to log the `Policy` as a pointer instead of a dereferenced pointer because that invoked the `Policy` `String()` method which is cleaner and does not print the secret details.

The `Policy` struct `String()` method is updated to remove newlines so that all the lines are written to the logs.

_producer/producer_protocol_handler.go_ had no instances of the `Proposal` or `Policy` objects being logged with the secret details in its `sendMessage()` function, but since _agreementbot/consumer_protocol_handler.go_ does, I thought it would be nice to put the same fix in both for symmetry and as a preventative measure.

### DIFF
--- a/agreementbot/consumer_protocol_handler.go
+++ b/agreementbot/consumer_protocol_handler.go
@@ -139,9 +139,25 @@ func (w *BaseConsumerProtocolHandler) sendMessage(mt interface{}, pay []byte) er
 		return errors.New(fmt.Sprintf("input message target is %T, expecting exchange.MessageTarget", mt))
 	}
 
+	logMsg := string(pay)
+
+	// Try to demarshal pay into Proposal struct
+	if glog.V(5) {
+		if newProp, err := abstractprotocol.DemarshalProposal(logMsg); err == nil {
+			// check if log message is a byte-encoded Proposal struct
+			if len(newProp.AgreementId()) > 0 {
+				// if it is a proposal, obscure the secrets
+				if logMsg, err = abstractprotocol.ObscureProposalSecret(logMsg); err != nil {
+					// something went wrong, send empty string to ensure secret protection
+					logMsg = ""
+				}
+			}
+		} 
+	}
+
 	// Grab the exchange ID of the message receiver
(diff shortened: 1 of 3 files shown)


agreementbot/consumer_protocol_handler.go
```diff
@@ -139,9 +139,37 @@ func (w *BaseConsumerProtocolHandler) sendMessage(mt interface{}, pay []byte) er
 		return errors.New(fmt.Sprintf("input message target is %T, expecting exchange.MessageTarget", mt))
 	}
 
+	logMsg := string(pay)
```
dabooz, line 142: 1) All of this needs to happen inside a check for V(5) logging being on or off
2) Why are you demarshalling as a proposal and as a policy? I dont think we ever send just a policy

policy/policy_file.go
```diff
@@ -503,42 +503,46 @@ func (self *Policy) Is_Version(v string) bool {
 
 func (self *Policy) String() string {
 	res := ""
-	res += fmt.Sprintf("Name: %v Version: %v, Pattern: %v\n", self.Header.Name, self.Header.Version, self.PatternId)
-	res += "API Specifications\n"
+	res += fmt.Sprintf("Name: %v, Version: %v, Pattern: %v, ", self.Header.Name, self.Header.Version, self.PatternId)
```
dabooz, line 506: getting rid of the new lines really makes this long log entry very hard to read

dabooz (changes requested):

policy/policy_file.go
Polber: It was always a long log entry because it was not calling the .String() method to begin with. Now with this change, it does call .String(), but if the newlines are in there, it won't print anything after the first newline occurrence.

agreementbot/consumer_protocol_handler.go
Polber: I did it in case it is ever sent in the future. I suppose it could hinder performance though, so I will change it to just check for a proposal.

dabooz (approved):

PR #2696, opened 2021-07-30
Issue 2693 - Remote e2edev test failure

MaxMcAdam:
Signed-off-by: Max McAdam <max@fredcom.com>


MaxMcAdam:
The agbot did not send a proposal for the weather service when the node is running sall pattern because of this error
`Compatibility Error: Properties [] do not satisfy Node constraint  [iame2edev == true NONS==false || NOGPS == false || NOLOC == false || NOPWS == false || NOHELLO == false || NOK8S == false]. Underlying error: The required properties iame2edev==true were not found in the available properties `
The wi.producerpolicy is set to the node policy at the beginning of InitiateNewAgreement. The weather pattern has no required services so the api spec list given to GetMergedProducerPolicyForPattern was empty and it returned a nil pointer. Then wi.producerpolicy wasn't changed and InitiateAgreement failed because the node policy constraints are not met by the consumerpolicy properties.

dabooz (approved):

PR #2697, opened 2021-07-30
Issue 2623 - Add hzn commands for new SDO API's

Polber:


test/gov/hzn_sdo.sh
```diff
@@ -0,0 +1,1166 @@
...
+rc=$?
+if [[ $rc -eq 7 && "$cmdOutput" == *'Error:'*'Could not get'*'HZN_SDO_SVC_URL'* ]]; then
+	echo -e "${PREFIX} received expected error response."
+else
+	echo -e "${PREFIX} Failed: Wrong error response from 'hzn sdo keys create' without HZN_SDO_SVC_URL set: exit code: $rc, output: $cmdOutput."
+	exit 1
+fi
+export HZN_SDO_SVC_URL=$HZN_SDO_SVC_URL_SAVE
+
+echo -e "${PREFIX} Testing 'hzn sdo keys create <key-file>'"
```
bmpotter, line 710: Do you need to remove this key before trying to create it, in case the script exited last time after the create and before the remove? (This is only necessary in the e2edev remote case.)

test/gov/hzn_sdo.sh
```diff
@@ -0,0 +1,1166 @@
...
+		echo -e "${PREFIX} Failed: 'hzn sdo voucher download <device-id> -f /tmp': downloaded file does not match expected output."
+		exit 1
+	fi
+else
+	echo -e "${PREFIX} Failed: Wrong error response from 'hzn sdo voucher download': exit code: $rc, output: $cmdOutput."
+	exit 1
+fi
+rm /tmp/$VOUCHER_DEVICE_ID.json
+
+# echo -e "${PREFIX} Testing 'hzn sdo voucher download' with incorrect voucher name argument"
```
bmpotter, line 500: Why is this test case commented out?

test/gov/hzn_sdo.sh
```diff
@@ -0,0 +1,1166 @@
...
+cmdOutput=$(hzn sdo keys remove test-sdo-key 2>&1)
+rc=$?
+if [[ $rc -eq 0 && "$cmdOutput" == *'Key'*'successfully deleted from the SDO owner services.'* ]]; then
+	echo -e "${PREFIX} completed."
+else
+	echo -e "${PREFIX} Failed: Wrong error response from 'hzn sdo keys remove': exit code: $rc, output: $cmdOutput."
+	exit 1
+fi
+
+# echo -e "${PREFIX} Testing 'hzn sdo keys remove <keyName>' when key does not exist"
```
bmpotter, line 817: Why is this test case commented out?

bmpotter (changes requested):

test/gov/hzn_sdo.sh
Polber: The keys are now removed before creation

test/gov/hzn_sdo.sh
Polber: I did that because the API updates were not reflected in the test environment which would cause these tests to fail. They have now been updated and uncommented.

test/gov/hzn_sdo.sh
Polber: I did that because the API updates were not reflected in the test environment which would cause these tests to fail. They have now been updated and uncommented.

bmpotter (approved):

dabooz (approved):

PR #2709, opened 2021-08-04
Issue 2420 - More work on node's service configstate changes

sirotsinskuy:


linggao:
I further studies the original code and realized that the configstate change already uses the exchange version as the base. So no need to store a local copy. I will close (cancel) this PR and issue 2420. Thanks a lot for your effort.

PR #2783, opened 2021-08-20
Issue 2782 - e2edev failure on service secret tests

MaxMcAdam:
Signed-off-by: Max McAdam <max@fredcom.com>


MaxMcAdam:
This addresses 2 issues, one with the test and one with the code.
The test issue is that sometimes the netspeed service is returned from curling /services before the container has started so the test cannot find the container and fails.
The code issue is that ResourcesCreate only runs once for singleton services so services that use the singleton but start later do not get their agreement id added to the secret. Then when the original service using the singleton stops the secret is removed. There is a test specifically for this that only fails intermittently since it depends on the order the services started.

dabooz (approved):

PR #2801, opened 2021-08-26
Issue 2800 - Add logic to check for unsupported sdo key names

Polber:
Signed-off-by: polber <jeff@thekinards.com>


cli/sdo/key.go
```diff
@@ -287,3 +297,25 @@ func checkEmptyKeyFields(key KeyFile) error {
...
+        if (r < 'a' || r > 'z') && (r < '0' || r > '9') && r != '-' {
+            return errors.New("Key name can only contain lowercase characters, numbers, and hyphens")
+        }
+    }
+	return nil
+}
+
+func checkKeyCountry(s string) error {
+	err := "Country name must be a 2 Letter Country Code"
+	if len(s) > 2 {
```
bmpotter, line 312: My understanding is that the country code must be exactly 2 chars, so just check `len(s) == 2`

test/gov/hzn_sdo.sh
```diff
@@ -67,6 +67,30 @@ cat <<'EOF' > /tmp/sdo_key2.json
 }
 EOF
 
+cat <<'EOF' > /tmp/sdo_key_unsupported.json
```
bmpotter, line 70: Why is this metadata unsupported and why don't i see it used anywhere?

bmpotter (changes requested):

test/gov/hzn_sdo.sh
Polber: Because I used the wrong name in the test. Thanks for the catch.

cli/sdo/key.go
Polber: It should be `len(s) != 2`. I'm not sure when or why I changed it to `>` but that is definitely incorrect.

bmpotter (dismissed):

bmpotter (approved):

PR #2898, opened 2021-10-06
Issue 2879 - assess each object manager policy

dlarson04:

- Look at each object manager policy individually
- Avoid long log messages printing destinations
- Batch destination updates to CSS to avoid a single huge POST
- Agbot can determine if an agent is already in the destination list to avoid unnecessary call to CSS

### DIFF
--- a/agreementbot/policy.go
+++ b/agreementbot/policy.go
@@ -15,24 +15,58 @@ import (
 	"strings"
 )
 
+const DESTINATIONS_LOG_LIMIT = 50
+
 // This function is called when an object is ready to be deployed to a node. It will perform the policy compatibility test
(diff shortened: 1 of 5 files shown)


agreementbot/policy.go
```diff
@@ -299,6 +349,31 @@ func (w *BaseAgreementWorker) HandleMMSObjectPolicy(cph ConsumerProtocolHandler,
 					}
 				}
 			}
+
+			glog.V(3).Infof(opLogstring(fmt.Sprintf("doug... batch size.... %v", CSSDestinationBatchSize)))
```
LiilyZhang, line 353: Can you remove "doug" in the log?

dabooz (approved):

PR #2956, opened 2021-12-15
Issue 2955 - Eliminate 1 call from agent

dlarson04:


agreement/agreement.go
```diff
@@ -430,6 +424,13 @@ func (w *AgreementWorker) CommandHandler(command worker.Command) bool {
 			deleteMessage = false
 		} else if pDevice, err := persistence.FindExchangeDevice(w.db); err != nil {
 			glog.Errorf(logString(fmt.Sprintf("unable to get device from the local database. %v", err)))
+		} else if there, err := w.messageInExchange(exchangeMsg.MsgId); err != nil {
```
dabooz, line 427: I would rather see this new code before the FindExchangeDevice call.

dabooz (changes requested):

agreement/agreement.go
dlarson04: Changed per request

dabooz (approved):

PR #3043, opened 2022-03-01
Issue 2996 - AgentAutoUpgrade: k8s cronjob to monitor cluster agent and start rollback process

Polber:
Signed-off-by: Jeff Kinard <jeff@thekinards.com>


agent-install/agent-uninstall.sh
```diff
@@ -365,6 +366,10 @@ function deleteAgentResources() {
     $KUBECTL delete pvc $PVC_NAME -n $AGENT_NAMESPACE
     set -e
 
+    log_info "Deleting auto-upgrade cronjob..."
+    $KUBECTL delete cronjob $AUTO_UPGRADE_CRONJOB_NAME -n $AGENT_NAMESPACE
+    set -e
+
     log_info "Deleting clusterrolebinding..."
```
linggao, line 373: What happens if the cron job does not exist?  (For backward compatibility)

LiilyZhang (changes requested):

LiilyZhang (approved):

PR #4235, opened 2025-01-31
Revert "Use ubi-micro instead of ubi-minimal to reduce the threat surface attack area."

LiilyZhang:
Reverts open-horizon/anax#4228


naphelps (approved):
LGTM

omordyk (approved):

PR #4304, opened 2025-04-04
Fix palmctl uninstall command output 

Yummikko:
## Description
When following the official gw deletion scenario, the gateway deletion automatically remove the node, causing an error output:
```
[ERROR] Error: node 'node-name' not found in org 'org-name'
[ERROR] command terminated with exit code 8
```
This message was confusing, as it implied the node "was not removed." With this change, exit code 8 is now interpreted as a successful outcome, indicating that the node was already removed.

### DIFF
--- a/agent-install/agent-uninstall.sh
+++ b/agent-install/agent-uninstall.sh
@@ -298,11 +298,15 @@ function deleteNodeFromManagementHub() {
     local node_id=$1
 
     log_info "Deleting node ${node_id} from the management hub..."
-
     set +e
-    $KUBECTL exec ${POD_ID} -n ${AGENT_NAMESPACE} -c "anax" -- bash -c "${EXPORT_EX_USER_AUTH_CMD}; hzn exchange node remove ${node_id} -f"
+    $KUBECTL exec ${POD_ID} -n ${AGENT_NAMESPACE} -c "anax" -- bash -c "${EXPORT_EX_USER_AUTH_CMD}; hzn exchange node remove ${node_id} -f" >/dev/null 2>&1  
+    local delete_exit_code=$?
     set -e
 
+    if [ $delete_exit_code -ne 0 ] && [ $delete_exit_code -ne 8 ]; then
+        log_warning "Node ${node_id} removal failed with exit code $delete_exit_code"
+    fi
+    
     log_debug "deleteNodeFromManagementHub() end"
 }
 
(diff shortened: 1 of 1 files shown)


agent-install/agent-uninstall.sh
```diff
@@ -317,10 +317,15 @@ function verifyNodeRemovedFromManagementHub() {
 
     set +e
     $KUBECTL exec ${POD_ID} -n ${AGENT_NAMESPACE} -c "anax" -- bash -c "${EXPORT_EX_USER_AUTH_CMD}; hzn exchange node list ${node_id}" >/dev/null 2>&1
-    if [ $? -ne 8 ]; then
-	    log_warning "Node was not removed from the management hub"
-    fi
+    local exit_code=$?
     set -e
```
kroczi, line 321: I do not think that this call is producing the error. based on the logs order it looks like it comes from line 303 in `deleteNodeFromManagementHub()`.

kroczi (changes requested):

agent-install/agent-uninstall.sh
```diff
@@ -242,14 +242,10 @@ function removeNodeFromLocalAndManagementHub() {
         fi
     else
         log_info "Node is not registered, skip unregister..."
-        if [[ "$DELETE_EX_NODE" == "true" ]]; then
-            log_info "Remove node from the management hub..."
```
kroczi, line 246: reinstate the log

agent-install/agent-uninstall.sh
```diff
@@ -242,14 +242,10 @@ function removeNodeFromLocalAndManagementHub() {
...
-            deleteNodeFromManagementHub $NODE_ID
-        fi
     fi
 
-    if [[ -n $NODE_STATE ]] && [[ "$DELETE_EX_NODE" == "true" ]]; then
-        verifyNodeRemovedFromManagementHub $NODE_ID
+    if [[ "$DELETE_EX_NODE" == "true" ]]; then
+        deleteAndVerifyNodeFromManagementHub $NODE_ID
     fi
```
kroczi, line 249: I'd call this function just `deleteNodeFromManagementHub`

agent-install/agent-uninstall.sh
```diff
@@ -290,38 +286,40 @@ function getEscapedExchangeUserAuth() {
...
-    log_debug "verifyNodeRemovedFromManagementHub() begin"
-
-    escaped_USER_AUTH=$(getEscapedExchangeUserAuth)
-    EXPORT_EX_USER_AUTH_CMD="export HZN_EXCHANGE_USER_AUTH=${escaped_USER_AUTH}"
-    local node_id=$1
-
-    log_info "Verifying node ${node_id} is removed from the management hub..."
+    if [ $delete_exit_code -eq 0 ] || [ $delete_exit_code -eq 8 ]; then
+        log_info "Deletion: Node ${node_id} has been successfully removed"
+    else
```
kroczi, line 305: I'd remove this info log, until we are sure the node was removed (after verification below)

kroczi (changes requested):

agent-install/agent-uninstall.sh
```diff
@@ -242,14 +242,11 @@ function removeNodeFromLocalAndManagementHub() {
         fi
     else
         log_info "Node is not registered, skip unregister..."
-        if [[ "$DELETE_EX_NODE" == "true" ]]; then
-            log_info "Remove node from the management hub..."
-            deleteNodeFromManagementHub $NODE_ID
-        fi
     fi
```
kroczi, line 246: You have moved the `deleteNodeFromManagementHub` out of the `else` statement. Was it intentional?

kroczi: I see that `unregister()` function does try to remove node it the flag `DELETE_EX_NODE` is true. So I am not sure if this change is going to work properly in all scenarios.

dlarson04 (changes requested):

PR #4311, opened 2025-04-10
Issue open-horizon#4312 - Update Ubuntu Version in GitHub Actions Workflows

Yummikko:
### Description 
Update Ubuntu 20.04 to a supported version. 

### References
[4312](https://github.com/open-horizon/anax/issues/4312])

### DIFF
--- a/.github/workflows/E2E-test.yml
+++ b/.github/workflows/E2E-test.yml
@@ -14,7 +14,7 @@ on:
 jobs:
   test:
     # The type of runner that the job will run on
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
 
     # Strategy allows specifying matrix axis(es) that will run for the test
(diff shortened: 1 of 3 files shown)


.github/workflows/E2E-test.yml
```diff
@@ -14,7 +14,7 @@ on:
 jobs:
   test:
     # The type of runner that the job will run on
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
```
LiilyZhang, line 17: @bencourliss is it a good practice to use `ubuntu-latest`? Should we use a concrete ubuntu version?

kroczi (approved):

dlarson04 (approved):
Some other components already use ubuntu-latest so this is good to be consistent.

Issue #4337, opened 2025-05-07
CWE-338: Weak Pseudo-Random


Yummikko:
### CWE Description
When a non-cryptographic PRNG is used in a cryptographic context, it can expose the cryptography to certain types of attacks. Often a pseudo-random number generator (PRNG) is not designed for cryptography. Sometimes a mediocre source of randomness is sufficient or preferable for algorithms that use random numbers. Weak generators generally take less processing power and/or do not use the precious, finite, entropy sources on a system. While such PRNGs might have very useful features, these same features could be used to break the cryptography

PR #4339, opened 2025-05-12
Issue open-horizon#4337 - Fix CVE-338: Weak Pseudo-Random

Yummikko:
### CWE Description

When a non-cryptographic PRNG is used in a cryptographic context, it can expose the cryptography to certain types of attacks. Often a pseudo-random number generator (PRNG) is not designed for cryptography. Sometimes a mediocre source of randomness is sufficient or preferable for algorithms that use random numbers. Weak generators generally take less processing power and/or do not use the precious, finite, entropy sources on a system. While such PRNGs might have very useful features, these same features could be used to break the cryptography

### References 
[4337](https://github.com/open-horizon/anax/issues/4337)

### DIFF
--- a/cutil/cutil.go
+++ b/cutil/cutil.go
@@ -3,7 +3,7 @@ package cutil
 import (
 	"bufio"
 	"crypto/md5"
-	"crypto/rand"
+	crypto_rand "crypto/rand"
 	"encoding/base64"
 	"encoding/hex"
 	"encoding/json"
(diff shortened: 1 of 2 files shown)


.github/workflows/E2E-test.yml
```diff
@@ -14,7 +14,7 @@ on:
 jobs:
   test:
     # The type of runner that the job will run on
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
```
omordyk, line 17: use latest

omordyk:
@LiilyZhang pls review

omordyk (approved):
LGTM

cutil/cutil.go
```diff
@@ -93,7 +93,7 @@ func GenerateRandomNodeId() (string, error) {
 
 	bytes := make([]byte, 20, 20)
 	nodeIdString := ""
-	_, err := rand.Read(bytes)
+	_, err := crypto_rand.Read(bytes)
```
LiilyZhang, line 96: what does this acturally changed? Isn't it still use `crypto/rand` library? @Yummikko

PR #4534, opened 2026-08-21
Docs issue 708 Fix TOC plus liquid warnings

### LINKED ISSUE #708: Automate setup of new edge node
There are approximately 15 steps for a horizon user to do to an edge node, once the OS is installed, to get the node setup, registered with horizon, and verified. These steps can be automated to make the onboarding process quicker and less error prone. This is not intended for the production use cases rob has, but before most companies adopt a technology they take it for a test drive, which in our case could involve setting up a few edge nodes manually and trying one of the existing edge applications. This is the place the edge node setup automation could be useful.

But it is tricky to automate processes like this correctly, because covering up underlying steps can become problematic when something goes wrong, or the user needs more flexibility. So we must be intentional about minimizing these pitfalls.

David Martin wrote a [good prototype of automating the horizon node setup](https://github.ibm.com/dcmartin/hzn-device-init), which we can use as a starting point, but it needs the following improvements:

- [ ] Restructure the code to make proper use of bash functions
- [ ] Provide Usage info
- [ ] Check all requirements up front, before doing anything: OS type/version, arch, input, env vars, etc.
- [ ] Be transparent, but succinct, about what the script is doing for the user, echoing certain commands where appropriate
- [ ] Enhance error handling to recognize even more problems
- [ ] Make the script idempotent, i.e. if an error occurs part way thru, once the user has fixed the problem they can simply run the script again and it will recognize what has already be done successfully.
- [ ] Make registering the pattern be optional (but the default)
- [ ] Have real command line parsing, so flags can be supported
- [ ] Remove loops looking for the firmware, images, and configstate values in `hzn node list`. A successful exit code from `hzn register` already assures these.

FYI, there are approx 15 steps that can be automated in https://console.test.cloud.ibm.com/docs/services/edge-fabric/adding-devices.html . Most of them **can't** be automated in the bluehorizon deb pkg because:

1) it's the steps for configuring apt and installing the deb pkgs, or
2) the steps require horizon user credentials (no way to give those to a deb pkg being installed), or
3) some steps are optional like starting the cpu2msghub example to verify the whole stack (no way to choose optional behavior in a deb install), or
4) installation of optional software (kafka is only required if using an edge service that sends data to kafka, so we wouldn't want bluehorizon to require always)

David Martin also wrote a script to create lots of devices: https://github.ibm.com/dcmartin/hzn-device-init and opened an issue: https://github.ibm.com/Edge-Fabric/project-mgmt/issues/131

Part of epic https://github.ibm.com/Edge-Fabric/project-summary/issues/35

euconnor:
## Description

Putting one topic into its correct location in the TOC, per docs issue 708. Also fix 2 liquid warnings.

Fixes #708 (partially)

## Type of change

Please delete options that are not relevant.

- [ ] This change requires a documentation update

This change **is** a documentation update.

## How Has This Been Tested?

Build help system locally and verified liquid warnings stopped appearing. Verified topic in correct location in local build.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -3,8 +3,8 @@ copyright: Contributors to the Open Horizon project
 years: 2022 - 2026
 title: Edge node agents (anax)
 description: Open Horizon Edge Node Agent Documentation
-lastupdated: 2026-04-08
-nav_order: 5
+lastupdated: 2026-08-21
+nav_order: 7
 has_children: True
 has_toc: False
 layout: page
--- a/docs/installing_ocp_edge_cluster_agent.md
+++ b/docs/installing_ocp_edge_cluster_agent.md
@@ -3,7 +3,7 @@ copyright: Contributors to the Open Horizon project
 years: 2022 - 2026
 title: Installing the agent on Red Hat OpenShift Container Platform Kubernetes edge cluster
 description: Documentation for installing the agent on OpenShift edge clusters
-lastupdated: 2026-04-07
+lastupdated: 2026-08-21
 nav_order: 1
 parent: Installing the agent on an edge cluster
 grand_parent: Edge node agents (anax)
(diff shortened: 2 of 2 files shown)


joewxboy (approved):
LGTM

